PR #2, opened 2018-10-16
Add Generic Build Script

jamiebrynes7:
We use a script similar to this all over the place - making a single generic one to reduce script duplication.

Once this is merged - will refactor the existing CI scripts to invoke this one


scripts/build.sh
```diff
@@ -0,0 +1,32 @@
...
+LOG_FILE=${4:-} # Optional argument - will log to console otherwise
+
+RUN_UNITY_PATH="$(dirname "$0")/tools/RunUnity/RunUnity.csproj"
+
+markStartOfBlock "Building ${WORKER_TYPE} for ${BUILD_TARGET}"
+
+pushd "${UNITY_PROJECT_DIR}"
+    dotnet run -p "${RUN_UNITY_PATH}" -- \
+        -projectPath "${UNITY_PROJECT_DIR}" \
+        -batchmode \
```
gkassabli, line 26: do we want -nographics here as well?

gkassabli (approved):

scripts/build.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+# This is a generic build script for the GDK for Unity.
+# The expected usage is as follows:
+#   bash build.sh <unity_project_dir> <worker_type> <build_target> <log_file (optional)>
+
+set -e -u -x -o pipefail
+
+source "$(dirname "$0")/scripts/pinned-tools.sh"
+source "$(dirname "$0")/scripts/profiling.sh"
```
gkassabli: I think /scripts/ should not be here - dirname will point to the scripts directory

scripts/build.sh
jamiebrynes7: I don't think so - it can cause problems with building I think. cc @zeroZshadow

zeroZshadow: It can screw up builds that do light baking and other stuff

jared-improbable (approved):

scripts/build.sh
```diff
@@ -0,0 +1,34 @@
...
+#   bash build.sh <unity_project_dir> <worker_type> <build_target> <log_file (optional)>
+
+set -e -u -x -o pipefail
+
+cd "$(dirname "$0")/../"
+
+source "scripts/pinned-tools.sh"
+source "scripts/profiling.sh"
+
+UNITY_PROJECT_DIR=$1
```
jamiebrynes7, line 9: This might be relative to the working directory when this script was invoked.

By adding `cd "$(dirname "$0")/../"` back in - this will break

gkassabli: Not having cd... did break RUN_UNITY_PATH - dotnet complained about missing project.
Let me play with script to see what I can do to support relative path as well

jamiebrynes7: Sure - we probably want it to handle relative paths properly. It should be possible given:

1. We can get the path of this script
2. We can get the working directory it was invoked in

gkassabli: Ok, this version is verified to work on both relative and absolute paths. @jamiebrynes7 what do you think of it?

jamiebrynes7: Yeah that looks good to me :)

PR #4, opened 2018-10-22
Add Android dependencies for Unity

DanailPenev:
A lightweight Unity project in the `tools/` dir that has a static method to set the JDK, Android SDK and Android NDK paths. These paths are fetched from environment variables.


tools/AndroidDependencies/ProjectSettings/AudioManager.asset
```diff
@@ -0,0 +1,17 @@
+%YAML 1.1
```
gkassabli: unity n00b question. Do we need all of these assets to use a few line script?

scripts/prepare-unity-mobile.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+### This script is used to setup Improbable's internal build machines.
+### If you don't work at Improbable, this may be interesting as a guide to what software versions we use for our
+### automation, but not much more than that.
+
+set -e -u -x -o pipefail
```
gkassabli, line 7: More general question - how this code is intended to be used?

tools/AndroidDependencies/ProjectSettings/AudioManager.asset
jessicafalk: unity creates them by default

DanailPenev: It's the minimum requirements for something to be considered a Unity project, AFAIK. Maybe @firtina-improbable has more context on that? I can try removing the assets to check if it will still work, though I did some experiments with removing things.

scripts/prepare-unity-mobile.sh
DanailPenev: In the GDK scripts we clone the `gdk-for-unity-shared-ci` repo and call these scripts from there.

gkassabli: Can we (and do we want to) alter the script so that it runs on initial project launch? If we can user can skip the process of setting it up manually

.gitignore
```diff
@@ -211,3 +211,8 @@ project.lock.json
 #####
 
 logs/
+
+tools/AndroidDependencies/*
```
ghost: Why are we going for an "ignore all by default and whitelist" approach rather than the other way around? Also it'd be better for there to be a gitignore within the AndroidDependencies directory instead

scripts/prepare-unity-mobile.sh
```diff
@@ -0,0 +1,29 @@
...
+
+    RUN_UNITY_PATH="$(pwd)/tools/RunUnity/RunUnity.csproj"
+popd
+
+markStartOfBlock "Setting Android dependencies"
+
+pushd "tools/AndroidDependencies"
+    dotnet run -p "${RUN_UNITY_PATH}" -- \
+        -projectPath "." \
+        -batchmode \
```
ghost, line 23: can also go for -nographics I think

tools/AndroidDependencies/Assets/UnityAndroidDependenciesSetter.cs
```diff
@@ -0,0 +1,26 @@
...
+    private static class EnvironmentVariables
+    {
+        public static readonly string JdkPath = Environment.GetEnvironmentVariable("JAVA_HOME");
+        public static readonly string AndroidSdkPath = Environment.GetEnvironmentVariable("ANDROID_SDK_ROOT");
+        public static readonly string AndroidNdkPath = Environment.GetEnvironmentVariable("ANDROID_NDK_ROOT");
+    }
+
+    private static class UnityEditorPrefKeys
+    {
+        public const string JdkKey = "JdkPath";
```
ghost: "Key" suffix may not be necessary

tools/AndroidDependencies/ProjectSettings/AudioManager.asset
ghost: For a unity project, these files are good to add to source control.

But, why do we need a new unity project for this? Can't we add them to the main repo instead?

DanailPenev: They are CI specific.

scripts/prepare-unity-mobile.sh
DanailPenev: This requires the users to set up environment variables manually which I feel is more troublesome than simply adding it with the Unity GUI.

.gitignore
DanailPenev: Can move that there.

scripts/prepare-unity-mobile.sh
DanailPenev: We don't do `-nographics` for the build either. Better to stay consistent.

scripts/prepare-unity-mobile.sh
gkassabli: Fair enough, let's keep it limited to CI

tools/AndroidDependencies/ProjectSettings/AudioManager.asset
gkassabli: Should we create more generic project for CI? Or is it better to keep it tied to Android for now?

ghost: even if they are CI specific, it adds too much overhead vs. a script within other repo that only gets used in CI

ghost: or, can add it to the test project I guess?

scripts/prepare-unity-mobile.sh
ghost: There's a reason for the build to be with `-nographics` as it affects lighting builds, but that's not necessary here. However it does not add much benefit so it's OK for it not to be in here.

tools/AndroidDependencies/ProjectSettings/AudioManager.asset
DanailPenev: I know it's not ideal, but considering it's CI only I think this is a nice separation of concerns. It wouldn't make sense to me to have it in the Test Project or the GDK project as it's not related to either of these. It's literally setting up Unity preferences. If people strongly believe it shouldn't be here, I can move it but I think this is the most logical place, albeit the unnecessary overhead.

jamiebrynes7: I agree this doesn't belong in the test project. Is there no other way of changing Unity preferences? If not, we should definitely raise a post/issue as this is a wacky workflow. 

Also we aren't the only ones who have to build mobile clients in CI - so I'm surprised that this is how its done

ghost: I see, imo that's OK - this project is fine

DanailPenev:
The assets in the Unity project were actually not necessary. This should be the least amount of needed files for this to run. 

@jamiebrynes7 unfortunately Untiy doesn't support any other way of setting these programmatically. We can flag with them but this has been the case for a long time now (2+ years) and am not sure how interested they will be.

gkassabli (approved):
lgtm

tools/AndroidDependencies/.gitignore
```diff
@@ -0,0 +1,3 @@
+Library/
+Packages/
```
ghost: Is `Packages/` necessary to add to gitignore?
Or, the other way around: do you need to add `ProjectSettings` to gitignore?
For the `!*.meta` line, why is it necessary to have that line?

jessicafalk: good points!
both Packages and meta probably shouldn't be in that gitignore?

However, Library, Temp and obj should be.

tools/AndroidDependencies/Assets/UnityAndroidDependenciesSetter.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
```
jessicafalk: should we move this to the `Editor` folder so that it will automatically be excluded for builds? 
Sure, people are not supposed to build it, but if we ever extend that project for any reason

tools/AndroidDependencies/.gitignore
ghost: It depends how far we want to have in the "ignore all the things" slider:

baseline would be `Library`, `Temp`, and `obj`
next level would add ProjectSettings (minus project version) and Packages
maximum would be "everything (including meta files) but this .cs file"

tools/AndroidDependencies/Assets/UnityAndroidDependenciesSetter.cs
ghost: good idea

tools/AndroidDependencies/.gitignore
jessicafalk: Packages not really, right? because if we actually want to add Packages, we need them synchronized via git

ghost: For this PR I don't think any packages would be necessary unless we may need to do some package configuration later on

jessicafalk: yes, we don't need it right now. But why ignore it, if it might be necessary later. I don't see a reason to ignore the folder :)

DanailPenev: The meta files are in the .gitignore because they are ignored by default from the top-level .gitignore (which I believe utilises some template).

I am not .gitignore-ing ProjectSettings because it only contains the ProjectVersion.txt file right now.

Temp and obj are not created by this project I believe but if we can all agree with ignoring `Library`, `Temp` and `obj` and un-ignoring `.meta` we can go with that.

ghost: > The meta files are in the .gitignore because they are ignored by default from the top-level .gitignore (which I believe utilises some template).

Which file specifically?

jessicafalk: they sometimes do created and sometimes don't. Maybe Firtina knows the rules on when they get created.

ghost: I'll do some experiments on this branch :)

DanailPenev: @firtina-improbable https://github.com/spatialos/gdk-for-unity-shared-ci/blob/master/.gitignore this one

DanailPenev: I guess we can just remove the ignore from there :)

ghost: ".meta" : Ah I see, sorry i was looking at the wrong repo's gitignore file 🤦‍♂️

ghost (approved):

tools/AndroidDependencies/Assets/Editor/UnityAndroidDependenciesSetter.cs
```diff
@@ -0,0 +1,39 @@
...
+        
+        EditorPrefs.SetString(UnityEditorPrefKeys.Jdk, EnvironmentVariables.JdkPath);
+        EditorPrefs.SetString(UnityEditorPrefKeys.AndroidSdk, EnvironmentVariables.AndroidSdkPath);
+        EditorPrefs.SetString(UnityEditorPrefKeys.AndroidNdk, EnvironmentVariables.AndroidNdkPath);
+        
+        Console.WriteLine("Android preferences after execution:");
+        printPreferences();
+    }
+
+    private static void printPreferences()
```
jessicafalk: that should be capitalized. will do it when testing it

tools/AndroidDependencies/Assets/Editor/UnityAndroidDependenciesSetter.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        Console.WriteLine($"JDK Path: {EditorPrefs.GetString(UnityEditorPrefKeys.Jdk)}");
+        Console.WriteLine($"Android SDK Path: {EditorPrefs.GetString(UnityEditorPrefKeys.AndroidSdk)}");
+        Console.WriteLine($"Android NDK Path: {EditorPrefs.GetString(UnityEditorPrefKeys.AndroidNdk)}\n");
```
jessicafalk: no new line at the end needed, right?

PR #5, opened 2018-10-25
Fix building on OSX if Unity’s within /Applications/Unity/Unity.app

ghost:
If `/Applications/Unity/Unity.app` exists, we can use that one.

Q: do we / should we validate it matches projectVersion?


jessicafalk:
Can we validate?

ghost:
Depends... it's not done for Windows, so I can add "validate project version match" as a feature in this/another PR

DanailPenev:
If it's not done for Windows either, I would say validating is a job for a different PR.

jessicafalk:
agreed

jamiebrynes7:
Validating the Unity version is surprisingly hard. There's no `--version` switch.

tools/UnityPaths/UnityPaths.cs
```diff
@@ -53,6 +53,11 @@ public static string TryGetUnityPath()
                 return hubPath;
             }
 
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) && Directory.Exists(Path.Combine(DefaultUnityPath, "Unity.App")))
```
gkassabli, line 56: I just installed Unity with the Hub on my Mac and the path it is installed to is /Applications/Unity/Hub/Editor/_version_/Unity.app
Do we want to cover that scenario?

jessicafalk:
yeah, that's what I thought

tools/UnityPaths/UnityPaths.cs
jessicafalk: that is already covered afaik

gkassabli: Okay, reading the code, I guess that should be covered by the code above. If that is the case, ignore my comment

ghost: Aye, if it's the right version it should

jessicafalk (approved):

DanailPenev (approved):

PR #7, opened 2018-10-29
Add Test Runner

jamiebrynes7:
**TL;DR: Seeing this**
 ```
if [ $EDITMODE_TEST_RESULT -ne 0 ] || \
   [ $PLAYMODE_TEST_RESULT -ne 0 ] || \
   [ $CODE_GENERATOR_TEST_RESULT -ne 0 ] || \
   [ $TEST_PROJECT_EDITMODE_TEST_RESULT -ne 0 ] 
then
    >&2 echo "Tests failed! See above for more information."
    exit 1
fi
``` 
**made me really sad and was very error prone when editing so I made this instead.**

Resurrected from the old `experimental/meta-test-runner` branch from the `gdk-for-unity` repo. Adding a meta test runner to replace the contents of the test.sh script with a nicer way of adding new test sources and types of tests.

Currently the test script has to capture all the exit codes and or them all at the end to give a final result. This abstracts it away such that you have a single config file like:
```
{
    "modules": [
        {
            "type": "dotnet",
            "name": "Code Generator Tests",
            "parameters": {
                "project_path": "../workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/GdkCodeGenerator.csproj"
            }
        },
        {
            "type": "unity",
            "name": "Unity GDK Editmode Tests",
            "parameters": {
                "project_path": "../workers/unity/",
                "test_platform": "editmode"
            }
        },
        {
            "type": "unity",
            "name": "Unity GDK Playmode Tests",
            "parameters": {
                "project_path": "../workers/unity",
                "test_platform": "playmode"
            }
        },
        {
            "type": "unity",
            "name": "Test Project Editmode Tests",
            "parameters": {
                "project_path": "../test-project",
                "test_platform": "editmode"
            }
        }
    ]
}
```

And the test runner takes care of the rest. New test types can be easily added to the test runner as it has a modular structure.

### CODE REVIEW

File: tools/GdkTestRunner/Modules/ModuleLibrary.cs [hunk 1]
```diff
@@ -0,0 +1,60 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
```
jessicafalk, line 4: not needed

tools/GdkTestRunner/Model/BaseModuleDefinition.cs
```diff
@@ -0,0 +1,12 @@
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
```
jessicafalk, line 2: not needed

tools/GdkTestRunner/Modules/UnityModule.cs
```diff
@@ -0,0 +1,166 @@
+using System;
+using System.Diagnostics;
```
jessicafalk, line 2: not needed

tools/GdkTestRunner/Modules/UnityModule.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        protected override void AfterRun()
+        {
+            CleanLibraryAndTempFolders();
+        }
+
+        private void CleanLibraryAndTempFolders()
+        {
+            var libraryFolder = Path.Combine(unityProjectPath, "Library");
+            var tempFolder = Path.Combine(unityProjectPath, "Temp");
```
jessicafalk, line 125: should we also delete the obj folder?

tools/GdkTestRunner/Modules/UnityModule.cs
```diff
@@ -0,0 +1,166 @@
...
+        {
+            CleanLibraryAndTempFolders();
+        }
+
+        protected override bool InternalRun()
+        {
+            var currentDirectory = Environment.CurrentDirectory;
+            Environment.CurrentDirectory = unityProjectPath;
+
+            var unityPath = GetUnityExePath(Paths.TryGetUnityPath());
```
jessicafalk, line 80: would it be better to use the `RunUnity` project here to avoid duplicating code?

tools/GdkTestRunner/Modules/ModuleLibrary.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            foreach (var moduleType in moduleTypes)
+            {
+                var identifierAttribute = Attribute.GetCustomAttribute(moduleType,
+                    typeof(TestModuleIdentifierAttribute)) as TestModuleIdentifierAttribute;
+
+                if (identifierAttribute == null)
+                {
+                    LogManager.GetCurrentClassLogger().Error($"Encountered subclass of {typeof(BaseModule).FullName} " +
+                        $"without {typeof(TestModuleIdentifierAttribute).FullName}");
```
jessicafalk, line 31: what's the difference between `typeof(TestModuleIdentifierAttribute).FullName` and `nameof(TestModuleIdentifierAttribute)`?

tools/GdkTestRunner/Modules/UnityModule.cs
jamiebrynes7: Well its a binary so I'd have to invoke `dotnet run RunUnity` to invoke Unity which is another layer of abstraction

tools/GdkTestRunner/Modules/ModuleLibrary.cs
jamiebrynes7: Not sure - I think `nameof` might be compile-time

tools/GdkTestRunner/Modules/UnityModule.cs
jamiebrynes7: Linter failed me 😭

tools/GdkTestRunner/Modules/UnityModule.cs
jamiebrynes7: This was pulled from what we currently do in `test.sh`. Not sure what the implications of deleteing one and not the other is

tools/GdkTestRunner/Modules/UnityModule.cs
jessicafalk: but you wouldn't duplicate code :P

tools/GdkTestRunner/Modules/UnityModule.cs
jessicafalk: would delete obj as well just to make sure. they are all temporary folders that might appear and should be cleaned up

tools/GdkTestRunner/Modules/ModuleLibrary.cs
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            return (BaseModule) Activator.CreateInstance(moduleType, options, jsonContext);
+        }
+
+        public static string GetModuleDocumentation(string identifier)
+        {
+            if (!moduleDocumentation.TryGetValue(identifier, out var docs))
+            {
+                throw new ArgumentException($"Could not find a module with identifier: {identifier}");
```
jessicafalk, line 54: very small nit: one time you wrote `Cannot ... ` and here you use `Could not ... ` ;)

jared-improbable:
I think this is potentially a neat idea, but only once we have > 2 different types of test runners.
Are we projecting that we're going to have a variety of different non-unity tests to run? Since there's only one type right now, the architecture that you have isn't really tested "in anger" and may not actually be the best approach.

I suggest that you pare this down to do exactly one thing - invoke Unity and validate the specific inputs and outputs. If/when we need to abstract away multiple test runners, we can revisit and possibly resurrect this approach, if it is indeed the best fit solution.

PR #9, opened 2018-11-13
Check for all playback engines 

paulbalaji:
After checking for a Unity installation, check for all expected playback engines

Also includes drive-by fix for adding blank project to README 

Relates to: [UTY-1437](https://improbableio.atlassian.net/browse/UTY-1437)


scripts/prepare-unity.sh
```diff
@@ -37,11 +37,22 @@ export GOPATH
...
-  exit 0
 else
   echo "Could not find a Unity $UNITY_PINNED_VERSION installation."
 fi
 
+if [ -d "$IMPROBABLE_UNITY_VERSIONED/PlaybackEngines/LinuxStandaloneSupport" ] || \
+   [ -d "$IMPROBABLE_UNITY_VERSIONED/PlaybackEngines/AndroidPlayer" ] || \
+   [ -d "$IMPROBABLE_UNITY_VERSIONED/PlaybackEngines/iOSSupport" ] || \
+   [ -d "$IMPROBABLE_UNITY_VERSIONED/PlaybackEngines/WindowsStandaloneSupport" ] || \
+   [ -d "$IMPROBABLE_UNITY_VERSIONED/PlaybackEngines/MacStandaloneSupport" ]
```
ghost: For Mac this path is different for MacStandaloneSupport, also the directories should be:

`"$IMPROBABLE_UNITY_VERSIONED/Editor/Data/PlaybackEngines/..."`

paulbalaji: yes you're right - the paths here are the mac ones

will add the windows/mac changes 👍

scripts/prepare-unity.sh
```diff
@@ -37,11 +37,22 @@ export GOPATH
...
+then
+  echo "All PlaybackEngines are already installed in: $IMPROBABLE_UNITY_VERSIONED/PlaybackEngines."
+  exit 0
+else
+  echo "Could not find all playback engines."
```
jamiebrynes7: What happens if we get here because of a missing playback engine and try to reinstall all of Unity?

paulbalaji: then it re-installs all of Unity with all the modules specified above - did a test run on an agent and it appears to reinstall fine

scripts/prepare-unity.sh
```diff
@@ -41,13 +41,19 @@ else
...
+  PLAYBACK_ENGINE_ROOT="$IMPROBABLE_UNITY_VERSIONED/Editor/Data/PlaybackEngines"
+elif isMacOS; then
+  PLAYBACK_ENGINE_ROOT="$IMPROBABLE_UNITY_VERSIONED/PlaybackEngines"
+fi
+
+if [ -d "$PLAYBACK_ENGINE_ROOT/LinuxStandaloneSupport" ] && \
+   [ -d "$PLAYBACK_ENGINE_ROOT/AndroidPlayer" ] && \
+   [ -d "$PLAYBACK_ENGINE_ROOT/iOSSupport" ] && \
+   [ -d "$PLAYBACK_ENGINE_ROOT/WindowsStandaloneSupport" ] && \
+   [ -d "$PLAYBACK_ENGINE_ROOT/MacStandaloneSupport" ]
```
ghost: `    [ -d "$PLAYBACK_ENGINE_ROOT/MacStandaloneSupport" ] ` could still be invalid for Mac Editor, AFAIK it should be installed within the .app - unless our installer may do a redundant build support on top?

paulbalaji: yup - on mac the `MacStandaloneSupport` directory is inside the .app at `/Applications/Unity/Unity/Contents/PlaybackEngines/MacStandaloneSupport`

in which case I should probably not check for this directory when setting up it's a mac agent

scripts/prepare-unity.sh
```diff
@@ -37,11 +37,28 @@ export GOPATH
 
 if [ -d "$IMPROBABLE_UNITY_VERSIONED" ]; then
   echo "Unity is already installed in:" $IMPROBABLE_UNITY_VERSIONED
-  exit 0
 else
   echo "Could not find a Unity $UNITY_PINNED_VERSION installation."
 fi
 
+if isWindows; then
```
ghost: IMO these new additions can go within the `if [ -d "$IMPROBABLE_UNITY_VERSIONED" ]; then`

gkassabli (approved):
Lgtm in general, but restructuring code to avoid checking modules if Unity isn't installed would be nice

ghost (approved):

scripts/prepare-unity.sh
```diff
@@ -37,7 +37,24 @@ export GOPATH
...
+    PLAYBACK_ENGINE_ROOT="$IMPROBABLE_UNITY_VERSIONED/Editor/Data/PlaybackEngines"
+  elif isMacOS; then
+    PLAYBACK_ENGINE_ROOT="$IMPROBABLE_UNITY_VERSIONED/PlaybackEngines"
+  fi
+
+  if [ -d "$PLAYBACK_ENGINE_ROOT/LinuxStandaloneSupport" ] && \
+     [ -d "$PLAYBACK_ENGINE_ROOT/AndroidPlayer" ] && \
+     [ -d "$PLAYBACK_ENGINE_ROOT/iOSSupport" ] && \
+     [ -d "$PLAYBACK_ENGINE_ROOT/WindowsStandaloneSupport" ] && \
+     [ isMacOS || -d "$PLAYBACK_ENGINE_ROOT/MacStandaloneSupport" ]
```
jamiebrynes7, line 52: Why does Mac gotta be a special snowflake

jamiebrynes7 (approved):

scripts/prepare-unity.sh
gkassabli: Mac has it's own support installed in the Unity app package itself instead of standalone playback engine

jared-improbable: A comment to that effect would be very useful.

jared-improbable (approved):
Single comment aside, LGTM

scripts/prepare-unity.sh
paulbalaji: will add that 👍

PR #10, opened 2018-11-14
Fixing brackets when checking for mac support

paulbalaji:
`[` runs `test`, so had to separate `isMacOs` from the test


gkassabli (approved):
:check

jamiebrynes7 (approved):

PR #11, opened 2018-12-12
Adding Fastlane files

gkassabli:
This PR adds fastlane files that will be utilised to create distributable iOS builds

linked to: https://github.com/spatialos/gdk-for-unity/pull/651


fastlane/Fastfile
```diff
@@ -0,0 +1,31 @@
...
+#
+#     https://docs.fastlane.tools/actions
+#
+# For a list of all available plugins, check out
+#
+#     https://docs.fastlane.tools/plugins/available-plugins
+#
+
+# Uncomment the line if you want fastlane to automatically update itself
+# update_fastlane
```
gkassabli, line 14: All files are generated by fastlane. Fastfile is the only one that contains logic for now

fastlane/Fastfile
```diff
@@ -0,0 +1,31 @@
...
+
+default_platform(:ios)
+
+platform :ios do
+    desc "Description of what the lane does"
```
gkassabli: oops, will fix this

fastlane/report.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuites>
```
jessicafalk: probably shouldn't commit the report

fastlane/README.md
```diff
@@ -0,0 +1,29 @@
+fastlane documentation
```
jessicafalk: do we want this?
also, do we need to add some kind of license if we use fastlane?

paulbalaji:
you'll need to change the branch name so that TeamCity picks it up

fastlane/Fastfile
```diff
@@ -0,0 +1,30 @@
...
+platform :ios do
+    desc "Development build for iOS"
+    lane :dev do
+        cert(
+            development: true,
+            team_name: "Improbable Worlds Limited",
+        )
+        build_app(
+            scheme: "Unity-iPhone",
+            xcargs: "-allowProvisioningUpdates DEVELOPMENT_TEAM=\"27P8Y5FRY4\"",
```
jessicafalk, line 27: should we have that key here?

fastlane/Appfile
```diff
@@ -0,0 +1,6 @@
+# app_identifier("[[APP_IDENTIFIER]]") # The bundle identifier of your app
+# apple_id("[[APPLE_ID]]") # Your Apple email address
+
+
+# For more information about the Appfile, see:
+#     https://docs.fastlane.tools/advanced/#appfile
```
paulbalaji: is this meant to be indented like that?

fastlane/Fastfile
gkassabli: This is apple team ID, which isn't secret and shouldn't change

fastlane/Appfile
gkassabli: Generated code, let me update this

jessicafalk (approved):

PR #12, opened 2018-12-13
API Docs Generator

jamiebrynes7:
Okay this is _probably_ ready for a preliminary review.

I've tried to break it down into logical commits, but some parts reference future commits but it _should_ be fine to just look at the API call name and infer what it does. Otherwise it needs a better name!


tools/Docgen/Builder/DoxygenParser.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Xml.Serialization;
+using Docgen.Xml;
+
+namespace Docgen.Builder
+{
+    public static class DoxygenParser
+    {
+        private static List<string> BlackList = new List<string>
+        {
+            "index.xml"
```
jessicafalk, line 13: maybe add a short comment on why this needs to be blacklisted

tools/Docgen/Utils/Formatting.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// </summary>
+        /// <remarks>
+        ///     Note that the relative path has the target namespace stripped off the front of it.
+        /// </remarks>
+        /// <param name="compoundName">The Doxygen compound name.</param>
+        /// <returns>The relative path from the root of the API docs output folder.</returns>
+        public static string CompoundNameToRelativePath(string compoundName)
+        {
+            var targetNamespace = Options.Instance.TargetNamespace.Replace(".", "::") + "::";
+            var markdownSplitPath = compoundName.Replace(targetNamespace, "").Split("::")
```
jessicafalk, line 36: use string.Empty

tools/Docgen/Utils/Formatting.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+        
+        /// <summary>
+        ///     Selects the type name from a full qualified type name.
+        /// </summary>
+        /// <param name="fullyQualified">The fully qualified type name. Expected in the format of Foo.Bar.MyClass</param>
+        /// <returns>The type name.</returns>
+        public static string GetTypeFromFullyQualified(string fullyQualified)
+        {
+            return fullyQualified.Split(".").Reverse().First();
```
jessicafalk, line 73: why not just use Last()?

jamiebrynes7: Ah - yes

tools/Docgen/Builder/Builder.cs
```diff
@@ -0,0 +1,57 @@
...
+        public static void Build(DoxygenDatabase database)
+        {
+            CleanOutputDir(Options.Instance.OutputDirectory);
+            BuildDocs(database, Options.Instance.OutputDirectory);
+            BuildTableOfContents(database, Options.Instance.OutputDirectory);
+        }
+
+        private static void BuildTableOfContents(DoxygenDatabase database, string outputDirectory)
+        {
+            // Build the table of contents at {{outputDirectory}}/.toc/toc.md
```
jessicafalk, line 17: that is the toc just for the API docs right? how would this work together with the rest of the documentation?

tools/Docgen/Builder/Builder.cs
```diff
@@ -0,0 +1,57 @@
...
+            var filepath = Path.Combine(outputDirectory, ".toc", "toc.md");
+            Directory.CreateDirectory(Path.GetDirectoryName(filepath));
+            File.WriteAllText(filepath, new TableOfContentsGenerator(database).Generate());
+        }
+
+        private static void BuildDocs(DoxygenDatabase database, string outputDirectory)
+        {            
+            foreach (var details in database.TypeDetails)
```
jessicafalk, line 25: those loops could potentially be factored out into a method.

tools/Docgen/Builder/Builder.cs
jamiebrynes7: Yes and you can use the include macro in the main ToC.

tools/Docgen/Builder/Builder.cs
jamiebrynes7: Yeah agreed.

jamiebrynes7:
Moved to another repo

PR #13, opened 2018-12-19
Adding scripting backend option to Unity build script

jessicafalk:
So that we can support mono & il2cpp builds in ci. Depends on `feature/il2cpp-build` in the gdk repo.


scripts/build.sh
```diff
@@ -9,7 +9,8 @@ set -e -u -x -o pipefail
 UNITY_PROJECT_DIR=$1
 WORKER_TYPE=$2
 BUILD_TARGET=$3
-LOG_FILE=${4:-} # Optional argument - will log to console otherwise
+SCRIPTING_BACKEND=$4
```
gkassabli, line 12: should we add some default value to scripting backend?

gkassabli (approved):
lgtm

scripts/build.sh
jessicafalk: was considering it, but the defaults are different based on platform. So better to just require it

gkassabli: Fair enough!

paulbalaji (approved):

paulbalaji (changes requested):
need to update `prepare-unity.sh` to install the correct il2cpp modules

paulbalaji (approved):

ghost (approved):

scripts/prepare-unity.sh
```diff
@@ -45,11 +45,15 @@ if [ -d "$IMPROBABLE_UNITY_VERSIONED" ]; then
...
+  # Only check for Windows IL2CPP support on Windows agents, as Unity do not provide Windows-IL2CPP support for Mac.
+  # Only check for Mac IL2CPP support on Mac agents, as Unity do not provide Mac-IL2CPP support for Windows.
   if [ -d "$PLAYBACK_ENGINE_ROOT/LinuxStandaloneSupport" ] && \
      [ -d "$PLAYBACK_ENGINE_ROOT/AndroidPlayer" ] && \
      [ -d "$PLAYBACK_ENGINE_ROOT/iOSSupport" ] && \
      [ -d "$PLAYBACK_ENGINE_ROOT/WindowsStandaloneSupport" ] && \
-     ( isMacOS || [ -d "$PLAYBACK_ENGINE_ROOT/MacStandaloneSupport" ])
+     ( isMacOS || [ -d "$PLAYBACK_ENGINE_ROOT/MacStandaloneSupport" ]) && \
+     ( isMacOS || [ -d "$PLAYBACK_ENGINE_ROOT/WindowsStandaloneSupport/Variations/win64_development_il2cpp" ]) && \
+     ( isWindows || [ -d "$IMPROBABLE_UNITY_ROOT/Unity.app/Contents/PlaybackEngines/MacStandaloneSupport/Variations/macosx64_development_il2cpp" ])
```
paulbalaji: @gkassabli  just saw your previous comment

I see what you mean but then I think it becomes a lot less readable because atm it follows the "A -> B = ¬A || B" rule nicely

ghost (approved):

PR #15, opened 2019-01-17
Zip Script

jamiebrynes7:
This is the start of automating parts of our release process. This scripts zips up bundles for each of the repositories so you don't have to! 

If you know more about `tar` than I do, and know how to filter out hidden sub-directories, please let me know 😬!

This assumes that master of each repository is the release.


scripts/zip-projects.sh
```diff
@@ -0,0 +1,60 @@
...
+git clone git@github.com:spatialos/gdk-for-unity-fps-starter-project.git "${FPS_REPO}"
+
+pushd "${FPS_REPO}"
+    ./ci/bootstrap.sh
+    # Get rid of heavy & unnecessary folders before zipping it. Can't get tar to ignore hidden subdirectories sadly.
+    rm -rf ./.git
+    rm -rf ./.shared-ci
+    cd ..
+    pushd gdk-for-unity
+        rm -rf ./.git
```
paulbalaji, line 40: Should we also remove `./.github`, `./ci` generally?

scripts/zip-projects.sh
```diff
@@ -0,0 +1,60 @@
...
+# Step 2. FPS 
+FPS_REPO="${TEMP_DIR}/fps/gdk-for-unity-fps-starter-project"
+FPS_ZIP="$(pwd)/gdk-for-unity-fps-starter-project-${RELEASE_VERSION}.zip"
+git clone git@github.com:spatialos/gdk-for-unity-fps-starter-project.git "${FPS_REPO}"
+
+pushd "${FPS_REPO}"
+    ./ci/bootstrap.sh
+    # Get rid of heavy & unnecessary folders before zipping it. Can't get tar to ignore hidden subdirectories sadly.
+    rm -rf ./.git
+    rm -rf ./.shared-ci
```
paulbalaji, line 37: Probably don't need to zip `./scripts` either because we'll bundle gdk/fps together

scripts/zip-projects.sh
paulbalaji: Also `./test-project`?

scripts/zip-projects.sh
paulbalaji: same applies to the blank project

scripts/zip-projects.sh
```diff
@@ -0,0 +1,60 @@
...
+}
+
+trap cleanUp EXIT
+
+RELEASE_VERSION=$1
+
+TEMP_DIR=$(mktemp -d || mktemp -d -t tmp)
+
+# Step 1. GDK for Unity repo.
+CORE_REPO="${TEMP_DIR}/core/gdk-for-unity"
```
jessicafalk, line 20: git already does that for us, so either
- skip this step
- remove unnecessary files/folders here as well

scripts/zip-projects.sh
```diff
@@ -0,0 +1,60 @@
...
+CORE_ZIP="$(pwd)/gdk-for-unity-${RELEASE_VERSION}.zip"
+git clone git@github.com:spatialos/gdk-for-unity.git "${CORE_REPO}"
+
+pushd "${CORE_REPO}"
+    tar cf "${CORE_ZIP}" *
```
jessicafalk, line 25: you can exclude subdirectories with tar. tar is just veeery annoying and if not done right, it will just ignore  your parameters.
Here is an example of a "valid" command (and by valid I mean that it worked on my machine :P)

```
tar --exclude="workers" -cf test.zip *
```

- do not have a backslash at the end, it will ignore the exclude
- make sure to use relative paths for the exclude. I could not get it to work with absolute paths. the path for the zip can be absolute

jessicafalk: however, for some reason I was not able to unzip my zip without using tar... might need to check that. Windows said that it's an invalid zip

scripts/zip-projects.sh
jamiebrynes7: I was just removing the heavier folders (I.e. `.git` is ~200 MB on the gdk repo).

We could have a whole blacklist thing

scripts/zip-projects.sh
jamiebrynes7: The idea was to have consistency across the repos - but we could just omit the bundle on the core repo given git does have a source zip.

Also the `*` operator actually globs the local directory, which automatically omits any hidden files/directories - so its unnecessary to remove the folders here as well.

scripts/zip-projects.sh
jamiebrynes7: 🤔 That's odd - the zips opened fine for me. Although.. I was using 7Zip which probably supports far more compression schemes than the Windows one. 

Will test this and replace `tar` calls with `zip` calls instead. 

Or given this is might get complex, might just rewrite this into C#

scripts/zip-projects.sh
jamiebrynes7: Similar comment as to above

jared-improbable:
Why `tar` instead of `zip?

jamiebrynes7:
`zip` doesn't appear to be installed by puppet (at least on my machine!), but `tar` is. So I went with that one 🤷‍♂️ 

On reflection I might spin this out into a mini-packaging C# CLI tool so we can have proper file filtering.

scripts/zip-projects.sh
gkassabli: Would it be better if we extract the code into a single procedure and add use blacklist in it?

jamiebrynes7: Yes - but I'd rather not do that in bash, will probably rewrite this script in C#

PR #16, opened 2019-01-21
Packaging Tool

jamiebrynes7:
Expanding on from #15 - I've rewritten this into C# to make it more flexible. It looks for a file like `packer.config.json` it its local directory to describe how to package something. For example the `fps-starter-project` repo config would be:

```json
{
    "package_name": "gdk-for-unity-fps-starter-project",
    "version": "0.1.4",
    "git_packages": [
        {
            "clone_url": "git@github.com:spatialos/gdk-for-unity.git",
            "clone_dir" : "gdk-for-unity",
            "exclude_paths": [
                ".git",
                "ci",
                ".github"
            ]
        },
        {
            "clone_dir": "gdk-for-unity-fps-starter-project",
            "clone_url": "git@github.com:spatialos/gdk-for-unity-fps-starter-project.git",
            "exclude_paths": [
                ".git",
                "ci",
				"scripts",
                ".github"
            ]
        }
    ]
}
```

A fun thought I had is since the config is self describing you could just declare the repo that is a dependency and the packer could fetch the config from that repo. Fun times.

The intention is to run this in CI and upload packages automatically on releases. Ideally, the `version` field could be omitted from the config and inferred from another source.


tools/Packer/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+                var combinedPath = Path.Combine(repoPath, path);
+                Console.WriteLine($"Removing: {combinedPath}");
+                if (File.Exists(combinedPath))
+                {
+                    File.Delete(combinedPath);
+                }
+
+                if (Directory.Exists(combinedPath))
+                {
+                    //Directory.Delete(combinedPath, true);
```
zeroZshadow: Commented code

tools/Packer/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+        private static void CreatePackage(string temporaryDirectory, ConfigModel config)
+        {
+            var packageName = Path.Combine(Environment.CurrentDirectory, $"{config.CloneDir}-{config.Version}.zip");
+            ZipFile.CreateFromDirectory(temporaryDirectory, packageName);
+        }
+
+        private static void CloneRepo(string cloneUrl, string repoPath)
+        {
+            var arguments = string.Join(" ", new string[] { "clone", cloneUrl, $"\"{repoPath}\"" });
+            Console.WriteLine(arguments);
```
zeroZshadow: Should this debug print still be here?

tools/Packer/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+            using (var process = Process.Start("git", arguments))
```
zeroZshadow: This doesn't redirect output, so it will fail, but you wont know why

tools/Packer/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+                if (process.ExitCode != 0)
+                {
+                    throw new Exception($"Process exited with a non-zero error code ({process.ExitCode})");
+                }
+            }
+        }
+
+
+        // Ensure that we can delete the files inside the directory by removing ReadOnly attribute from sub-files.
+        private static void DeleteDirectoryRecursive(string path)
```
zeroZshadow: scary

zeroZshadow (approved):

tools/Packer/Program.cs
jamiebrynes7: Fixed

tools/Packer/Program.cs
jamiebrynes7: Fixed

tools/Packer/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        private static string GetTempDirectory()
+        {
+            var tempDirPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+            Directory.CreateDirectory(tempDirPath);
+            return tempDirPath;
+        }
+
+
```
paulbalaji: nit: extra line

paulbalaji (approved):

gkassabli:
I actually like the idea of dependencies declared by url, though current version is fine for our purposes atm

gkassabli (approved):

tools/Packer/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+                var combinedPath = Path.Combine(repoPath, path);
+                Console.WriteLine($"Removing: {combinedPath}");
+                if (File.Exists(combinedPath))
+                {
+                    File.Delete(combinedPath);
+                }
+
+                if (Directory.Exists(combinedPath))
+                {
+                    DeleteDirectoryRecursive(combinedPath);
```
jessicafalk: `Directory.Delete(combinedPath, true)` does the same

jessicafalk: talked offline, please add a comment on why we did not use it

tools/Packer/ConfigModel.cs
```diff
@@ -0,0 +1,21 @@
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Packer
+{
+    public class ConfigModel
+    {
+        [JsonProperty("clone_dir")] public string CloneDir;
```
jared-improbable: Since we're always writing this sort of thing, it may be worth thinking about using SnakeCaseNamingStrategy to automatically go from `snek_kehs` to `PascalCase` without the repeated (and slightly annoying) annotations. Example: https://www.newtonsoft.com/json/help/html/NamingStrategySnakeCase.htm

tools/Packer/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+            CloneRepo(cloneUrl, repoPath);
+
+            foreach (var path in pathsToRemove)
+            {
+                var combinedPath = Path.Combine(repoPath, path);
+                Console.WriteLine($"Removing: {combinedPath}");
+                if (File.Exists(combinedPath))
+                {
+                    File.Delete(combinedPath);
+                }
```
jared-improbable, line 82: Could really be `else if` since it's one or the other.

tools/Packer/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+
+                if (Directory.Exists(combinedPath))
+                {
+                    DeleteDirectoryRecursive(combinedPath);
+                }
+            }
+        }
+
+        private static void CreatePackage(string temporaryDirectory, ConfigModel config)
```
jared-improbable: `tempDir` is probably fine

tools/Packer/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            var packageName = Path.Combine(Environment.CurrentDirectory, $"{config.CloneDir}-{config.Version}.zip");
+            ZipFile.CreateFromDirectory(temporaryDirectory, packageName);
+        }
+
+        private static void CloneRepo(string cloneUrl, string repoPath)
+        {
+            var arguments = string.Join(" ", new string[] { "clone", cloneUrl, $"\"{repoPath}\"" });
```
jared-improbable: I think the `string` part of `string[]` is redundant.

tools/Packer/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+                {
+                    CloneRepoAndClean(Path.Combine(tempDir, dependency.CloneIntoDir), dependency.CloneUrl,
+                        dependency.PathsToRemove);
+                }
+
+                CreatePackage(tempDir, config);
+            }
+            catch (Exception e)
+            {
+                Console.Error.WriteLine(e.Message);
```
jared-improbable, line 39: Nice, writing to StdErr as you'd expect :)

tools/Packer/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+                    throw new Exception($"Process exited with a non-zero error code ({process.ExitCode})");
+                }
+                
+                Console.WriteLine(processOutput.ToString());
+            }
+        }
+
+
+        // Ensure that we can delete the files inside the directory by removing ReadOnly attribute from sub-files.
+        private static void DeleteDirectoryRecursive(string path)
```
jared-improbable: Rather than destructively deleting undesired files, why not write the logic for `ZipFile.CreateFromDirectory` and allow it exclude a set of paths (that is, use the `ZipArchive` APIs and extension methods.)

tools/Packer/Program.cs
jamiebrynes7: I think you are correct

tools/Packer/ConfigModel.cs
jamiebrynes7: 🤔 If that works then it cuts down the boilerplate a ton

tools/Packer/Program.cs
jamiebrynes7: That would probably be the safer method

tools/Packer/Program.cs
jamiebrynes7: Nothing to see here..

tools/Packer/Program.cs
jamiebrynes7: But what if..

tools/Packer/Program.cs
jamiebrynes7: Implemented this instead!

tools/Packer/ConfigModel.cs
jamiebrynes7: Yeah it works nicely!

tools/Packer/ConfigModel.cs
```diff
@@ -0,0 +1,36 @@
...
+using Newtonsoft.Json.Serialization;
+
+namespace Packer
+{
+    public class ConfigModel
+    {
+        public string CloneDir;
+        public string CloneUrl;
+        public string Version;
+        public List<string> PathsToRemove = new List<string>();
```
jared-improbable: `ExcludeDirs`?

tools/Packer/ConfigModel.cs
```diff
@@ -0,0 +1,36 @@
...
+        public List<GitDependency> GitDependencies = new List<GitDependency>();
```
jared-improbable: It's kind of an odd structure that there's a top level "git dependency" and then nested ones. Why not put it all in a single list and require that there be at least one item in it?

tools/Packer/Program.cs
```diff
@@ -0,0 +1,172 @@
...
+            var packageName = Path.Combine(Environment.CurrentDirectory, $"{config.CloneDir}-{config.Version}.zip");
+
+            if (File.Exists(packageName))
+            {
+                File.Delete(packageName);
+            }
+
+            var archive = ZipFile.Open(packageName, ZipArchiveMode.Create);
+
+            var blackList = config.PathsToRemove
```
jared-improbable: See comment above about the awkwardly nested structure - this is an example of how the nesting plays out to the detriment of readability.

tools/Packer/Program.cs
```diff
@@ -0,0 +1,172 @@
...
+                if (blackList.Contains(file))
+                {
+                    continue;
+                }
+
+                // Strip away the base path (temporary directory) length + the leading slash
+                var entry = archive.CreateEntry(file.Remove(0, basePathLength));
+                using (var stream = entry.Open())
+                {
+                    stream.Write(File.ReadAllBytes(file));
```
jared-improbable, line 89: One thing that may be missing is maintaining file attributes, for example "+x" on shell scripts.

tools/Packer/ConfigModel.cs
jamiebrynes7: It could be specific files as well

tools/Packer/Program.cs
jamiebrynes7: I don't think we have any shell scripts being zipped up - but will look into file attributes

tools/Packer/ConfigModel.cs
jamiebrynes7: Refactored this to what you suggest

tools/Packer/ConfigModel.cs
jamiebrynes7: Renamed to `ExcludePaths`

tools/Packer/Program.cs
```diff
@@ -16,13 +16,16 @@ public static void Main(string[] args)
         {
             var config = ConfigModel.FromFile(Path.Combine(Environment.CurrentDirectory, ConfigFile));
 
+            if (config.GitPackages.Count == 0)
+            {
+                Console.Error.WriteLine("Packer requires at least one Git package defined. None were found.");
```
jared-improbable, line 21: Environment.Exit(1)?

tools/Packer/Program.cs
```diff
@@ -58,18 +61,15 @@ private static void CreatePackage(string tempDir, ConfigModel config)
...
+            var blackList = config.GitPackages.SelectMany(package =>
+                package.ExcludePaths.Select(subPath => Path.Combine(tempDir, package.CloneDir, subPath))).ToHashSet();
 
             AddToArchiveRecursive(archive, tempDir, blackList, tempDir.Length + 1);
         }
 
         private static void AddToArchiveRecursive(ZipArchive archive, string path, HashSet<string> blackList,
             int basePathLength)
-        {
+        {
```
jared-improbable: Whitespace

jared-improbable (approved):

PR #17, opened 2019-01-21
Add Upload Assembly Scripts

jamiebrynes7:
Add an upload script for uploading assemblies. Given a prefix, it will create a unique assembly for this commit hash.

I.e. - given `gdk_core` as a prefix, it would produce an assembly something like `gdk_core_1a2b3c4d`.

This assembly name can be inferred from any other script with the same commit meaning you have have a separate `upload` and `launch` scripts.

Again.. on the road to automating our release process.


paulbalaji (approved):

jessicafalk (approved):

scripts/upload-assemblies.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+# This is a generic script for uploading assemblies in the GDK for Unity.
```
jared-improbable, line 3: Add error handling/usage print if the user forgets the PREFIX.
Any requirements as to where this script should be run from?

jamiebrynes7: Good call.

And so long as its a Git repo in a SpatialOS project - will add this to the script summary

PR #18, opened 2019-01-23
Fix MacOS IL2CPP detection

gkassabli:
Previously we detected IL2CPP support on MacOS only in non-versioned destinations. This caused incorrect detections of missing IL2CPP support (i.e https://teamcity.corp.i8e.io/viewLog.html?buildId=1809447&buildTypeId=Develop_UnityGdk_BuildTestMac&tab=buildLog&_focus=595#_state=559,638)
This PR corrects IL2CPP detection to use needed versioned destination.


paulbalaji (approved):
nice catch 💯

jessicafalk (approved):

PR #19, opened 2019-01-23
Tweak Upload Script

jamiebrynes7:
After writing a script which used `upload-assemblies` (https://github.com/spatialos/gdk-for-unity/pull/704), it appeared that it would be easier and safer to refactor getting an assembly name out into a method and asking a user to pass in the full assembly name.


scripts/pinned-tools.sh
```diff
@@ -96,3 +96,17 @@ DOTNET_VERSION="$(dotnet --version)"
 if isWindows; then
   export MSBuildSDKsPath="${PROGRAMFILES}/dotnet/sdk/${DOTNET_VERSION}/Sdks"
 fi
+
+# Creates an assembly name based on a prefix and the current Git hash.
```
gkassabli: It might be a bit cleaner to mention using the argument instead of prefix

gkassabli (approved):
lgtm!

jessicafalk (approved):

PR #20, opened 2019-01-28
Fix Zips

jamiebrynes7:
This PR fixes the zips created by the tool to work with Windows + MacOS built in system unzipping tools.


oblm (approved):
**Tested:**
@jamiebrynes7  created a zip and opened it on both MacOS and Windows with the default unzipping tools.

gkassabli (approved):

PR #22, opened 2019-02-08
BuildKite support for shared_ci

gkassabli:
Files to support BuildKite runs on shared_ci repo.


gkassabli:
Buildkite is green!

.buildkite/premerge.definition.yaml
```diff
@@ -0,0 +1,11 @@
+agent_queue_id: trigger-pipelines
+description: Unity GDK Shared CI repository pre-merge CI pipeline.
```
jamiebrynes7: `GDK for Unity` 😉

jamiebrynes7 (approved):
LGTM module small change

.buildkite/premerge.definition.yaml
gkassabli: Oops, will fix this!

PR #23, opened 2019-02-18
Update AndroidDependencies to 2018.3.5f1

paulbalaji:
Update AndroidDependencies to 2018.3.5f1

Did this by opening `AndroidDependencies` in 2018.3.5f1


jessicafalk (approved):

zeroZshadow:
After this fix, we should check if this is still required.
I've done a local test and Unity seems totally capable to find the Android SDK by itself as long as the `ANDROID_SDK_ROOT` env variable is set. And the newer unity has an embedded JDK, so it no longer needs that path. According to older blogposts it also looks at the `JAVA_HOME` env var to find the JDK as well.

jared-improbable:
Shouldn't this go into `develop`?

jessicafalk:
we don't have a develop branch for this repo as it's not really versioned so far

PR #24, opened 2019-02-22
Release tool, Initial

andrew-improbable:
Hey all,

Here is the initial ReleaseTool.

There are two commands:

`prep` and `release` in `PrepCommand.cs` and `ReleaseCommand.cs` respectively.

Other points to note:

* I had to make changes to the `Packer` so that I could invoke the package logic from another project.
* The draft release title is just the version number eg. "0.5" instead of "GDK for Unity 0.5". This is because it was difficult to determine the release prefix (and it is more in-line with other github projects). We can revisit this (feel free to drop a comment if you feel strongly about it and I can see what I can do).
* We use a hybrid `GitClient` which uses some APIs from `libgit2sharp` and some calls directly to the shell.
* I still need to make sure it builds on CI properly.

Please leave comments!


tools/ReleaseTool/.gitignore
```diff
@@ -0,0 +1,140 @@
...
+[Rr]elease/
+x64/
+[Bb]in/
+[Oo]bj/
+
+# MSTest test Results
+[Tt]est[Rr]esult*/
+[Bb]uild[Ll]og.*
+
+*_i.c
```
jared-improbable: You can gut loads of things from this file

tools/ReleaseTool/App.config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
```
jared-improbable: Do you need this? It calls for .NETFramework, and the app is .NET Core

tools/ReleaseTool/CommandsCommon.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    static class CommandsCommon
+    {
+        private const string ReleaseBranchNameTemplate = "feature/release-{0}";
+
+        private const string GdkPinnedFilename = "gdk.pinned";
+
+        public static void UpdateGdkVersion(GitClient gitClient, string gdkVersion)
+        {
+            Console.WriteLine("Updating {0}...", GdkPinnedFilename);
```
jared-improbable: I'd generally steer towards using `$""` string interpolation rather than positional arguments these days.

tools/ReleaseTool/CommandsCommon.cs
```diff
@@ -0,0 +1,34 @@
...
+
+            if (!File.Exists(GdkPinnedFilename))
+            {
+                throw new InvalidOperationException(string.Format("Could not upgrade gdk version as the file, {0}, does not exist", GdkPinnedFilename));
```
jared-improbable: Use string interpolation

tools/ReleaseTool/CommandsCommon.cs
```diff
@@ -0,0 +1,34 @@
...
+            {
+                writer.WriteLine(gdkVersion);
+            }
+
+            gitClient.StageFile(GdkPinnedFilename, UriKind.Relative);
+        }
+
+        public static string ReleaseBranchName(string version)
+        {
+            return string.Format(ReleaseBranchNameTemplate, version);
```
jared-improbable, line 27: And again, interpolation

tools/ReleaseTool/Common.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using System.IO;
+
+namespace ReleaseTool
+{
+    internal static class Common
+    {
+        public static string GetCurrentDirectory()
```
jared-improbable: This seems like a dubious function to have around, unless you're not using `Path.Combine`?

tools/ReleaseTool/Common.cs
```diff
@@ -0,0 +1,52 @@
...
+                return originalPath;
+            }
+
+            var relativePath = new Uri(originalPath.Substring(2), UriKind.Relative);
+            var homePath = new Uri(GetHomePath(), UriKind.Absolute);
+
+            return new Uri(homePath, relativePath).AbsolutePath;
+        }
+
+        private static string AppendDirectorySeparator(string originalPath)
```
jared-improbable: Again, this is very suspicious.

tools/ReleaseTool/Common.cs
```diff
@@ -0,0 +1,52 @@
...
+            {
+                return originalPath + altDirectorySeparator;
+            }
+
+            return originalPath + directorySeparator;
+        }
+
+        private static string GetHomePath()
+        {
+            return AppendDirectorySeparator((Environment.OSVersion.Platform == PlatformID.Unix ||
```
jared-improbable: Does this https://docs.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=netcore-2.2 not work cross-platform?

tools/ReleaseTool/GitClient.cs
```diff
@@ -0,0 +1,179 @@
...
+            var pushArguments = string.Format(PushArgumentsTemplate, options.GitRemote, remoteBranchName);
+            if (forcePush)
+            {
+                pushArguments += PushForceFlag;
+            }
+
+            RunGitCommand(pushArguments, "push branch");
+        }
+
+        private void RunGitCommand(string arguments, string description)
```
jared-improbable: I'd flip the order of these arguments

tools/ReleaseTool/GitClient.cs
```diff
@@ -0,0 +1,179 @@
...
+                {
+                    break;
+                }
+
+                bool shouldRetry = false;
+
+                while (true)
+                {
+                    Console.WriteLine(string.Format("Failed to {0}. Retry (y/n)?", description));
+                    string retryInput = Console.ReadLine();
```
jared-improbable: `.ToLower()`, simplify the comparisons below a bit

tools/ReleaseTool/GitClient.cs
```diff
@@ -0,0 +1,179 @@
...
+                Console.WriteLine("");
+
+                if (process.ExitCode == 0)
+                {
+                    break;
+                }
+
+                bool shouldRetry = false;
+
+                while (true)
```
jared-improbable, line 165: Is this tool always going to be run manually?

tools/ReleaseTool/GitClient.cs
```diff
@@ -0,0 +1,179 @@
...
+        private const string DefaultMasterBranch = "master";
+
+        private const string GitCommand = "git";
+        private const string PushArgumentsTemplate = "push {0} HEAD:refs/heads/{1}";
+        private const string FetchArguments = "fetch";
+        private const string PushForceFlag = " -f";
+
+        private const string RemoteBranchRefTemplate = "{1}/{0}";
+
+        public interface IGitOptions
```
jared-improbable, line 30: Why an interface?

tools/ReleaseTool/GitClient.cs
```diff
@@ -0,0 +1,179 @@
+﻿using System;
+using System.Diagnostics;
+using System.Security;
+using CommandLine;
+using LibGit2Sharp;
+using LibGit2Sharp.Handlers;
+
+namespace ReleaseTool
+{
+    internal class GitClient : IDisposable
```
jared-improbable, line 14: Some documentation explaining the hybrid approach here will be useful.

tools/ReleaseTool/GitClient.cs
```diff
@@ -0,0 +1,179 @@
...
+
+                process.StartInfo.UseShellExecute = false;
+
+                process.StartInfo.FileName = GitCommand;
+                process.StartInfo.Arguments = arguments;
+
+                process.Start();
+                process.WaitForExit();
+
+                Console.WriteLine("");
```
jared-improbable: Don't you want to see the git output?

tools/ReleaseTool/GitHubClient.cs
```diff
@@ -0,0 +1,97 @@
...
+            var matches = Regex.Match(remote, RemoteUrlRegex);
+
+            if (!matches.Success)
+            {
+                throw new ArgumentException(string.Format("Failed to parse remote {0}. Not a valid github repository.", remote));
+            }
+
+            var repositoryTask = octoClient.Repository.Get(matches.Groups[1].Value, matches.Groups[2].Value);
+
+            return repositoryTask.Result;
```
jared-improbable, line 69: `.Result` can cause deadlocks - prefer to `await` the result instead.

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -0,0 +1,284 @@
...
+        private readonly Options options;
+
+        public PrepCommand(Options options)
+        {
+            this.options = options;
+        }
+
+        public int Run()
+        {
+            var gitClient = new GitClient(options);
```
jared-improbable, line 75: Prefer to use `using(var gitClient)` instead

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -0,0 +1,284 @@
...
+            var gitHubClient = new GitHubClient(options);
+            try
+            {
+                if (gitClient.HasStagedOrModifiedFiles())
```
jared-improbable, line 79: This test makes it seem like the tool is intended to be run locally, shouldn't this happen via a CI step?

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -0,0 +1,284 @@
...
+            {
+                writer.Write(newChangeLog.ToString());
+            }
+
+            gitClient.StageFile(ChangeLogFilename, UriKind.Relative);
+        }
+
+        private void UpdatePackerConfig(GitClient gitClient)
+        {
+            // TODO:
```
jared-improbable: TODO

tools/ReleaseTool/packages.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
jared-improbable: You should be able to get rid of this file, since there are references in the project file now.

tools/ReleaseTool/ReleaseCommand.cs
```diff
@@ -0,0 +1,203 @@
...
+            {
+                while (!reader.EndOfStream)
+                {
+                    var line = reader.ReadLine();
+                    
+                    if (line.StartsWith("## "))
+                    {
+                        changedSection += 1;
+
+                        if (changedSection == 3)
```
jared-improbable, line 211: ??? Comments or examples, please.

tools/ReleaseTool/ReleaseCommand.cs
```diff
@@ -0,0 +1,203 @@
...
+                    {
+                        Console.WriteLine("Re-adding branch protection...");
+                        gitHubClient.AddAdminBranchEnforcement(gitHubRepo, options.MasterBranch);
+                    }
+                }
+                
+                // Create release
+                CreateRelease(gitHubClient, gitHubRepo, gitClient);
+
+                Console.WriteLine("");
```
jared-improbable: nit: prefer `string.Empty`

tools/ReleaseTool/ReleaseCommand.cs
```diff
@@ -0,0 +1,203 @@
...
+                    }
+                }
+
+                // Push to master
+                Console.WriteLine("Pushing to {0}...", options.MasterBranch);
+                Console.WriteLine("Removing branch protection...");
+                bool masterHasBranchEnforcement = gitHubClient.RemoveAdminBranchEnforcement(gitHubRepo, options.MasterBranch);
+                try
+                {
+                    gitClient.Push(options.MasterBranch, /* forcePush */ false);
```
jared-improbable: Alternative API suggestion: `gitClient.Push` and `gitClient.ForcePush`, make it really clear.

tools/ReleaseTool/GitHubClient.cs
```diff
@@ -0,0 +1,97 @@
...
+using System;
+using System.IO;
+using System.Text.RegularExpressions;
+using OctoClient = Octokit.GitHubClient;
+
+namespace ReleaseTool
+{
+    internal class GitHubClient
+    {
+        private const string DefaultKeyFileLocation = "~/.ssh/github.token";
```
jared-improbable, line 15: If this is running on CI, we should have a better method for delivering secrets.

jared-improbable (dismissed):
This seems to be setup to run from a local machine; even as a first draft that's problematic, and I'd like to see a refactor (and hopefully simplification) so that it can run from any build agent on CI.

tools/ReleaseTool/.gitignore
andrew-improbable: Yep can remove the entire file. This was an artifact of copying from another repo! :P

tools/ReleaseTool/App.config
andrew-improbable: Yup can remove! :P

tools/ReleaseTool/CommandsCommon.cs
andrew-improbable: Will change all of these :)

tools/ReleaseTool/Common.cs
andrew-improbable: The problem is that the GitClient requires relative paths to add files.

And to calculate the relative paths we use URIs. But a URI without a trailing slash is considered a file not a directory. Will neaten this up.

tools/ReleaseTool/Common.cs
jessicafalk: the special folder enum does work accross different platforms afaik. It depends which folder you are looking for whether there is a mapping on both windows and mac

tools/ReleaseTool/GitClient.cs
andrew-improbable: See PrepCommand.Options (implements IGitOptions and IGitHubOptions).

tools/ReleaseTool/GitClient.cs
andrew-improbable: Done.

tools/ReleaseTool/GitClient.cs
andrew-improbable: Yes this already outputs the git output into the shell (by setting the UseShellExecute flag).

tools/ReleaseTool/GitClient.cs
andrew-improbable: Hmm good point, will add an option.

tools/ReleaseTool/GitHubClient.cs
andrew-improbable: can't as the method is not async.

tools/ReleaseTool/PrepCommand.cs
andrew-improbable: I don't do that because I catch exceptions here.

The `try { } finally { }` is just the decomposition of `using { }`.

tools/ReleaseTool/PrepCommand.cs
andrew-improbable: Yes I originally intended for this tool to be run locally, but it can also be put in a CI step if we want.

tools/ReleaseTool/packages.config
andrew-improbable: Done

tools/ReleaseTool/ReleaseCommand.cs
andrew-improbable: Done

tools/ReleaseTool/ReleaseCommand.cs
andrew-improbable: Changed all

tools/ReleaseTool/CommandsCommon.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            if (!File.Exists(GdkPinnedFilename))
+            {
+                throw new InvalidOperationException("Could not upgrade gdk version as the file, " +
+                                                    $"{GdkPinnedFilename}, does not exist");
+            }
+
+            using (var writer = new StreamWriter(GdkPinnedFilename))
+            {
+                writer.WriteLine(gdkVersion);
```
jamiebrynes7: Can just do: `File.WriteAllText(GdkPinnedFilename, gdkVersion)` rather than the using block

tools/ReleaseTool/CommandsCommon.cs
```diff
@@ -0,0 +1,35 @@
...
+            }
+
+            gitClient.StageFile(GdkPinnedFilename, UriKind.Relative);
+        }
+
+        public static string ReleaseBranchName(string version)
```
jamiebrynes7: Nit - rename to `GetReleaseBranchName`

tools/ReleaseTool/Common.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.IO;
+
+namespace ReleaseTool
+{
+    internal static class Common
+    {
+        public static string ReplaceHomePath(string originalPath)
```
jamiebrynes7, line 8: Why do we need to do this?

tools/ReleaseTool/GitHubClient.cs
jamiebrynes7: We would bake the secret into the agent during the `platform-bakery` stage. 

In which case we would probably do this (similar to how toolbelt OAuth token is stored).

We may want to look into rotating the secrets every so often

tools/ReleaseTool/GitHubClient.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        public GitHubClient(IGitHubOptions options)
+        {
+            this.octoClient = new OctoClient(ProductHeader);
+            this.options = options;
+        }
+            
+        public void LoadCredentials()
+        {
+            using (var reader = new StreamReader(Common.ReplaceHomePath(options.GitHubTokenFile)))
```
jamiebrynes7: Can you not do `File.ReadAllText()`?

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -0,0 +1,328 @@
...
+        public PrepCommand(Options options)
+        {
+            this.options = options;
+        }
+
+        public int Run()
+        {
+            if (string.IsNullOrEmpty(options.Version))
+            {
+                Console.Error.WriteLine("Release version was not specified.");
```
jamiebrynes7: Is this not a feature of the command line library? To mark an argument as required?

tools/ReleaseTool/PrepCommand.cs
jamiebrynes7: We will probably want to run this in CI at some point

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -0,0 +1,328 @@
...
+                PullRequest pullRequest;
+                try
+                {
+                    pullRequest = gitHubClient.CreatePullRequest(gitHubRepo, branchName, options.DevBranch,
+                        string.Format(PullRequestTemplate, options.Version));
+                }
+                catch (Exception e)
+                {
+                    Console.WriteLine("WARNING: Creating Pull Request Failed. This might be because the pull " +
+                                      $"request already exists? Error: {e}");
```
jamiebrynes7: Is this really a warning or is it an error?

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -0,0 +1,328 @@
...
+        }
+
+        private void UpdatePackageJson(string packageFile, GitClient gitClient)
+        {
+            Console.WriteLine($"Updating {packageFile}...");
+
+            JObject jsonObject;
+            using (var streamReader = new StreamReader(packageFile))
+            {
+                jsonObject = JObject.Parse(streamReader.ReadToEnd());
```
jamiebrynes7, line 156: Instead of doing a raw `JObject` - would it be better to define the JSON models as classes? 

https://www.newtonsoft.com/json/help/html/DeserializeObject.htm

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -0,0 +1,328 @@
...
+                        {
+                            dependencies[property.Name] = options.Version;
+                        }
+                    }
+                }
+            }
+
+            using (var streamWriter = new StreamWriter(packageFile))
+            {
+                streamWriter.Write(jsonObject.ToString());
```
jamiebrynes7: Again can do `File.WriteAllText()`

Also does this prettify the JSON?

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -0,0 +1,328 @@
...
+            if (!ShouldUpdateGdkVersion())
+            {
+                return;
+            }
+
+            CommandsCommon.UpdateGdkVersion(gitClient, options.GdkVersion);
+        }
+
+        /**
+         * Simple method for editing the ChangeLog
```
jamiebrynes7, line 194: Have you looked into using a Markdown library (https://github.com/lunet-io/markdig for example)?

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -0,0 +1,328 @@
...
+                    }
+
+                    newChangeLog.AppendLine(line);
+                }
+            }
+
+            using (var writer = new StreamWriter(ChangeLogFilename))
+            {
+                writer.Write(newChangeLog.ToString());
+            }
```
jamiebrynes7: `File.WriteAllText()` 😉

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -0,0 +1,328 @@
...
+        private void UpdatePackerConfig(GitClient gitClient)
+        {
+            if (!File.Exists(PackerConfigFile))
+            {
+                return;
+            }
+            
+            Console.WriteLine($"Updating {PackerConfigFile}...");
+
+            JObject jsonObject;
```
jamiebrynes7: Similar comment about using JSON models - in fact you can probably use the one that's already defined in the Packer csproj

tools/ReleaseTool/ReleaseCommand.cs
```diff
@@ -0,0 +1,216 @@
...
+    /// Runs the commands required for releasing a candidate.
+    /// * Does a final bump of the gdk.pinned (if needed).
+    /// * Merges the candidate branch into develop.
+    /// * Pushes develop to origin/develop and origin/master.
+    /// * Creates a GitHub release draft.
+    /// </summary>
+    internal class ReleaseCommand
+    {
+        private const string UpdateGdkCommitMessageTemplate = "Update to release commit for Gdk {0}.";
+        private const string MergeCommitMessageTempalte = "Merge release {0}.";
```
jamiebrynes7: Tempalte -> Template

tools/ReleaseTool/ReleaseCommand.cs
```diff
@@ -0,0 +1,216 @@
...
+            }
+            finally
+            {
+                gitClient.Dispose();
+            }
+
+            return 0;
+        }
+
+        private string BranchName()
```
jamiebrynes7: `GetBranchName()`

tools/ReleaseTool/ReleaseCommand.cs
```diff
@@ -0,0 +1,216 @@
...
+                gitClient.CheckoutRemoteBranch(branchName);
+
+                // Make Changes if necessary
+                UpdateGdkVersion(gitClient);
+
+                // Merge into develop
+                Console.WriteLine($"Merging {branchName} into {options.DevBranch}...");
+                var branchCommit = gitClient.GetHeadCommit();
+                gitClient.CheckoutRemoteBranch(options.DevBranch);
+                gitClient.Merge(branchCommit, string.Format(MergeCommitMessageTempalte, options.Version));
```
jamiebrynes7: Is this a squash merge?

jamiebrynes7: Also this bypasses our CI gates? 

Can this merge the open PR into develop instead?

tools/ReleaseTool/ReleaseCommand.cs
jamiebrynes7: I think using a Markdown library would help make these sections more legible

jamiebrynes7 (changes requested):
Consider looking into a logger setup (NLog for example) to replace the Console.WriteLine() calls everywhere

This will allow us to log verbosely to file and normally to stdout and can be configurable over the command line with a `-v` switch or similar.

Also generally you can use `File.ReadAllText()` or `File.WriteAllText()` instead of `StreamReader/Writer` objects. 

Would strongly recommend writing a JSON model for json files we are deserializing - it will make the code easier to read and more maintainable.

Also a quick look into a Markdown parsing library could help with the sections that parse and edit Markdown

tools/ReleaseTool/CommandsCommon.cs
andrew-improbable: Done

tools/ReleaseTool/CommandsCommon.cs
andrew-improbable: Done

tools/ReleaseTool/GitClient.cs
andrew-improbable: Done

tools/ReleaseTool/GitHubClient.cs
andrew-improbable: Changed

tools/ReleaseTool/PrepCommand.cs
andrew-improbable: True xD -- have changed

tools/ReleaseTool/PrepCommand.cs
andrew-improbable: I would, but this is a unity json, and I didn't really want to implement the entire manifest model (especially since they can add extra things). :(

tools/ReleaseTool/PrepCommand.cs
andrew-improbable: Changed -- ye it does prettify the JSON :)

tools/ReleaseTool/PrepCommand.cs
andrew-improbable: Yes I did. There is an AST in the library. But you cannot go from the AST to Markdown. You can only go from Markdown to the AST.

Also it was kind of nightmarish to edit the AST. I also had a look at other markdown libraries.

tools/ReleaseTool/PrepCommand.cs
andrew-improbable: Changed

tools/ReleaseTool/PrepCommand.cs
andrew-improbable: Yup agreed, I can definitely use that :P

tools/ReleaseTool/ReleaseCommand.cs
andrew-improbable: Fixed

tools/ReleaseTool/ReleaseCommand.cs
andrew-improbable: So the api for merging an open PR is a very finicky. (We would have to search through all PRs and look for the correct one).

There are a few options:
* We could leave as is and (unfortunately) this will bypass our merge gates -- but we could put build dependencies on BuildKite.
* We could make the PR web address an argument.
* We could search through all open PRs (potentially difficult).

Which do you prefer?

Willn also change this to a Squash Merge.

tools/ReleaseTool/ReleaseCommand.cs
andrew-improbable: Done

tools/ReleaseTool/ReleaseCommand.cs
andrew-improbable: (See comment above, the AST in markdig / other markdown libraries; is quite difficult to manipulate). For this particular use case it might easier, but its not possible for writing Markdown :(

tools/ReleaseTool/PrepCommand.cs
jamiebrynes7: You can implement a partial one - but not sure how that interacts with re-serializing objects 😕

tools/ReleaseTool/ReleaseCommand.cs
jamiebrynes7: Could you give a PR id as an argument? Rather than the full url?

andrew-improbable:
Addressed comments. PTAL.

tools/Packer/ConfigModel.cs
```diff
@@ -30,5 +30,10 @@ public static ConfigModel FromFile(string path)
 
             return JsonConvert.DeserializeObject<ConfigModel>(File.ReadAllText(path), settings);
         }
+
+        public void ToFile(string path)
```
jamiebrynes7, line 34: Nit: Rename `WriteToFile`

`ToFoo` is a convention that would usually have the following signature: `Foo ToFoo()` such that it does a conversion

tools/ReleaseTool/GitHubClient.cs
```diff
@@ -35,10 +35,7 @@ public GitHubClient(IGitHubOptions options)
             
         public void LoadCredentials()
         {
-            using (var reader = new StreamReader(Common.ReplaceHomePath(options.GitHubTokenFile)))
-            {
-                octoClient.Credentials = new Credentials(reader.ReadLine());
-            }
+            octoClient.Credentials = new Credentials(File.ReadAllText(Common.ReplaceHomePath(options.GitHubTokenFile)));
```
jamiebrynes7: Do you need to `ReplaceHomePath` here? Surely C#'s standard library can resolve `~` in a path?

jamiebrynes7 (commented):
I still think its worth adding https://nlog-project.org/ to this project since it does many things and logs a lot. 

It will give some clarity when looking at logs and will allow us to filter logs a bit better. (Debug vs Info)

andrew-improbable:
Ah ye sorry @jamiebrynes7 -- will add NLog forgot about that!

tools/ReleaseTool/GitHubClient.cs
andrew-improbable: It couldn't but I will dig deeper for a solution.

Either way this is only used to get the GitHub API key (which will be removed when we add a better secret sharing flow).

tools/ReleaseTool/GitClient.cs
```diff
@@ -35,6 +37,8 @@ public interface IGitOptions
 
             [Option("master-branch", Default = DefaultMasterBranch, HelpText = "The master branch.")]
             string MasterBranch { get; set; }
+            
+            bool Unattended { get; set; }
```
jamiebrynes7: Nit: `IsUnattended`

tools/ReleaseTool/GitClient.cs
```diff
@@ -122,20 +126,20 @@ public void ForcePush(string remoteBranchName)
...
         }
         
-        private static void RunGitCommand(string description, string arguments)
+        private void RunGitCommand(string description, string arguments)
         {
             while (true)
             {
-                Console.WriteLine($"Attempting to {description}. Running command [{GitCommand} {arguments}]");
-                Console.WriteLine(string.Empty);
+                Logger.Info("Attempting to {0}. Running command [{1} {2}]", description,
```
jamiebrynes7: Hmm would this be more appropriate as `debug` or `trace` logs?

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -48,6 +49,9 @@ public class Options: GitClient.IGitOptions, GitHubClient.IGitHubOptions
             [Option('d', "override-date", HelpText = "Override the date of the release. Leave blank to use the current date.")]
             public DateTime? OverrideDate { get; set; }
             
+            [Option('u', "unattended", HelpText = "Whether to run in unattended mode.")]
+            public bool Unattended { get; set; }
```
jamiebrynes7: Nit: `IsUnattended`

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -103,30 +107,16 @@ public int Run()
...
-                Console.WriteLine(string.Empty);
-                Console.WriteLine($"Release hash: {gitClient.GetHeadCommit().Sha}");
-                if (pullRequest != null)
-                {
-                    Console.WriteLine(string.Empty);
-                    Console.WriteLine($"Pull request available: {pullRequest.HtmlUrl}");
-                }
+                Logger.Info("Successfully created release!");
+                Logger.Info("Release hash: {0}", gitClient.GetHeadCommit().Sha);
+                Logger.Info("Pull request available: {0}", pullRequest.HtmlUrl);
```
jamiebrynes7, line 117: Could open the URL automatically 😉 

https://stackoverflow.com/questions/4580263/how-to-open-in-default-browser-in-c-sharp

tools/ReleaseTool/ReleaseCommand.cs
```diff
@@ -29,6 +31,9 @@ public class Options: GitClient.IGitOptions, GitHubClient.IGitHubOptions
 
             [Option('g', "update-gdk", HelpText = "The git hash of the version of the gdk to upgrade to. (Only if this is a project).")]
             public string GdkVersion { get; set; }
+            
+            [Option('u', "unattended", HelpText = "Whether to run in unattended mode.")]
+            public bool Unattended { get; set; }
```
jamiebrynes7: Nit: `IsUnattended`

tools/ReleaseTool/ReleaseCommand.cs
```diff
@@ -103,22 +108,21 @@ public int Run()
...
                 // Create release
                 var release = CreateRelease(gitHubClient, gitHubRepo, gitClient);
 
-                Console.WriteLine(string.Empty);
-                Console.WriteLine($"Release Complete! Release hash: {gitClient.GetHeadCommit().Sha}");
-                Console.WriteLine(string.Empty);
-                Console.WriteLine($"Draft release: {release.HtmlUrl}");
+                Logger.Info("Release Successful!");
+                Logger.Info("Release hash: {0}", gitClient.GetHeadCommit().Sha);
+                Logger.Info("Draft release: {0}", release.HtmlUrl);
```
jamiebrynes7, line 123: Similarly - if we are not in unattended mode, we could open the URL in the browser

tools/ReleaseTool/ReleaseCommand.cs
```diff
@@ -174,12 +181,12 @@ private static string GetReleaseNotesFromChangeLog()
         {
             if (!File.Exists(ChangeLogFilename))
             {
-                Console.WriteLine("Could not get draft release notes, as the change log file, " +
-                                  $"{ChangeLogFilename}, does not exist.");
+                Logger.Warn("Could not get draft release notes, as the change log file, " +
```
jamiebrynes7: Hmmm should this be an error rather than a warning? This is if the `CHANGELOG.md` file actually doesn't exist right?

tools/ReleaseTool/GitClient.cs
andrew-improbable: Fixed

tools/ReleaseTool/GitClient.cs
andrew-improbable: Fixed

tools/ReleaseTool/PrepCommand.cs
andrew-improbable: Done 👍

tools/ReleaseTool/ReleaseCommand.cs
andrew-improbable: Fixed

tools/ReleaseTool/ReleaseCommand.cs
andrew-improbable: Done! :)

tools/ReleaseTool/ReleaseCommand.cs
andrew-improbable: Changed! :)

tools/ReleaseTool/GitHubClient.cs
```diff
@@ -35,7 +39,15 @@ public GitHubClient(IGitHubOptions options)
...
         {
-            octoClient.Credentials = new Credentials(File.ReadAllText(Common.ReplaceHomePath(options.GitHubTokenFile)));
+            if (!string.IsNullOrEmpty(options.GitHubToken))
+            {
+                octoClient.Credentials = new Credentials(options.GitHubToken);
+            }
+            else
+            {
+                octoClient.Credentials = new Credentials(File.ReadAllText(
+                    Common.ReplaceHomePath(options.GitHubTokenFile)));
```
jamiebrynes7, line 54: Should check if this file exists

scripts/prepare-release.sh
```diff
@@ -0,0 +1,21 @@
...
+fi
+
+RELEASE_VERSION="$(buildkite-agent meta-data get \"release-version\")"
+GITHUB_SSK_KEY="$(imp-ci secrets read --environment=production --buildkite-org=improbable --secret-type=github-personal-access-token --secret-name=ci/improbable/unity-gdk/github-personal-access-token)"
+
+if [ -f "gdk.pinned" ]; then
+	GDK_VERSION="$(buildkite-agent meta-data get \"gdk-version\")"
+fi
+
+dotnet run -p $(dirname "$0")../tools/ReleaseTool/ReleaseTool.csproj -- prep --version=$RELEASE_VERSION --update-gdk=$GDK_VERSION --github-key=$GITHUB_SSK_KEY --unattended
```
jamiebrynes7: Nit: newline at the end

scripts/release.sh
```diff
@@ -0,0 +1,21 @@
...
+fi
+
+RELEASE_VERSION="$(buildkite-agent meta-data get \"release-version\")"
+GITHUB_SSK_KEY="$(imp-ci secrets read --environment=production --buildkite-org=improbable --secret-type=github-personal-access-token --secret-name=ci/improbable/unity-gdk/github-personal-access-token)"
+
+if [ -f "gdk.pinned" ]; then
+	GDK_VERSION="$(buildkite-agent meta-data get \"gdk-version\")"
+fi
+
+dotnet run -p $(dirname "$0")../tools/ReleaseTool/ReleaseTool.csproj -- release --version=$RELEASE_VERSION --update-gdk=$GDK_VERSION --github-key=$GITHUB_SSK_KEY --unattended
```
jamiebrynes7: Nit: new line at the end

scripts/prepare-release.sh
jamiebrynes7: Could we split parameters on lines? 

i.e. - 
```
dotnet run -p $(dirname "$0")../tools/ReleaseTool/ReleaseTool.csproj -- \
	prep \
	--version=$RELEASE_VERSION \
	--update-gdk=$GDK_VERSION \
...
```

jamiebrynes7 (approved):
Run the scripts through shellcheck.net 😉 

Also - will need to check if Git on the BK agents have users set up - otherwise commit and push operations will fail 

But other than that - few comments and then good 😄

PR #25, opened 2019-02-25
Move re-usable scripts to shared-ci

paulbalaji:
#### Description
- add `build-worker.sh` to replace individual project `build.sh` scripts
- add post-command hook
- add conditions for setting `RUN_UNITY_PATH`
- delete `upload-assemblies.sh` and made it a function in `pinned-tools.sh`

#### Tests
used this branch to run GDK/FPS pipelines

#### Documentation
n/a

#### Primary reviewers
If your change will take a long time to review, you can name at most two primary reviewers who are ultimately responsible for reviewing this request. @ mention them.


.buildkite/hooks/pre-exit
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -euo pipefail
+[[ -n "${DEBUG:-}" ]] && set -x
+
+# The premerge step runs on linux agents!
+if [[ ${BUILDKITE_AGENT_META_DATA_OS} == "windows" ]]; then
```
jessicafalk: the comment says it runs on linux agents, but here we are checking for windows?

scripts/build-worker.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+
+# This is a generic build script for the GDK for Unity.
+# The expected usage is as follows:
+#   bash build.sh <unity_project_dir> <worker_type> <build_target> <log_file (optional)>
```
jessicafalk: that comment seems slightly outdated ;)

scripts/build-worker.sh
```diff
@@ -0,0 +1,50 @@
...
+#   if [ -z ${UNITY_ASSET_CACHE_IP} ]; then
+#       ASSET_CACHE_ARG="-CacheServerIPAddress ${UNITY_ASSET_CACHE_IP}"
+#   else
+#       ASSET_CACHE_ARG=""
+#   fi
+
+ASSET_CACHE_ARG=${UNITY_ASSET_CACHE_IP:+-CacheServerIPAddress "${UNITY_ASSET_CACHE_IP}"}
+
+pushd "$(dirname "$0")/../"
+    source "scripts/pinned-tools.sh"
```
jessicafalk: I personally prefer to have all the `source` commands at the top to have some kind of structure in the scripts (I know you probably didn't make that change ;) )

scripts/pinned-tools.sh
```diff
@@ -111,6 +111,21 @@ function setAssemblyName() {
   ASSEMBLY_NAME="${1}_${GIT_HASH}"
 }
 
+# Uploads an assembly, given a $PREFIX and $PROJECT_NAME
+function uploadAssembly() {
+  if [ "$#" -ne 2 ]; then
+    echo "'uploadAssembly' expects two arguments."
+    echo "Example usage: 'uploadAssembly \$PREFIX \$PROJECT_NAME'"
```
jessicafalk: what does `\$PREFIX ` print? exactly that? (I honestly don't know)
though I think so far we have been using `<prefix>` to indicate which arguments are needed for a  command.

rename it to `<assembly prefix>` to make it clear what the prefix is for

scripts/build.sh
```diff
@@ -1,50 +1,28 @@
...
-popd
+  if [[ ${WORKER_TYPE} == "AndroidClient" ]]; then
+      scripts/prepare-unity-mobile.sh "$(pwd)/../logs/PrepareUnityMobile.log"
+  fi
 
-markStartOfBlock "Building ${WORKER_TYPE} for ${BUILD_TARGET} and ${SCRIPTING_BACKEND}"
+  if [[ ${WORKER_TYPE} == "iOSClient" ]]; then
+      if ! isMacOS; then
+          echo "I can't build for iOS!"
+          exit 0
```
jessicafalk: do we expect this to be run on non-macos machines?

scripts/build-worker.sh
```diff
@@ -0,0 +1,50 @@
...
+    source "scripts/profiling.sh"
+
+    RUN_UNITY_PATH="$(pwd)/tools/RunUnity/RunUnity.csproj"
+popd
+
+markStartOfBlock "Building ${WORKER_TYPE} for ${BUILD_TARGET} and ${SCRIPTING_BACKEND}"
```
jessicafalk: are we still using them in buildkite?

.buildkite/hooks/pre-exit
paulbalaji: Just copying across @jamiebrynes7 's PR: https://github.com/spatialos/gdk-for-unity/pull/744

jamiebrynes7: We only want to run the purge script on Windows - surprisingly the `premerge` step runs on a Linux box (the step which uploads the rest of the steps)

scripts/build.sh
jamiebrynes7: Yes - this is just a guard against misconfiguration, in which case this may want to exit 1? 

I've left the build iOS on the steps list and this early outs - once we have Mac Buildkite we can target that step at the Mac agent

jessicafalk: yeah, I think we should exit 1 here to make it clear that something got misconfigured

.buildkite/hooks/pre-exit
jessicafalk: maybe add a comment that we want to run the purge script on windows to make it a bit less confusing

.buildkite/hooks/post-command
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
jamiebrynes7, line 1: Why is this file here now?

scripts/build.sh
```diff
@@ -1,50 +1,28 @@
 #!/usr/bin/env bash
```
jamiebrynes7: This change will affect TeamCity as the `build-test` script uses this

scripts/pinned-tools.sh
```diff
@@ -104,13 +104,28 @@ function setAssemblyName() {
...
   fi
 
   ASSEMBLY_NAME="${1}_${GIT_HASH}"
 }
 
+# Uploads an assembly, given an assembly prefix and a project name
+function uploadAssembly() {
+  if [ "$#" -ne 2 ]; then
+    echo "'uploadAssembly' expects two arguments."
+    echo "Example usage: 'uploadAssembly <assembly prefix> <project name>'"
```
jamiebrynes7, line 118: Do we need to support the project name argument?

scripts/purge.ps1
```diff
@@ -0,0 +1,33 @@
+# This script is used to kill all dangling procceses at the end of a Buildkite build.
```
jamiebrynes7, line 1: This script is duplicated here and in `ci/purge.ps1`

jamiebrynes7 (changes requested):

.buildkite/hooks/post-command
jamiebrynes7: Disregard this

scripts/purge.ps1
```diff
@@ -4,9 +4,7 @@
 # to find all processes of a particular name which have file locks on our build (project) directory.
 # We can then attempt to kill each of these processes if any exist.
 
-$ScriptLocation = $MyInvocation.MyCommand.Path
-$ScriptDir = Split-Path $ScriptLocation
-cd "$ScriptDir/../.."
+cd $projectRoot
```
jamiebrynes7, line 10: Small nit - can we add usage to the comment above since it takes an argument now :)

scripts/pinned-tools.sh
paulbalaji: yes, this is to defend against the case where the `spatialos.json` refers to a different project than the deployment launcher defaults. controlling the project end-to-end gives us guarantee that if this step succeeds that the assembly has been uploaded to the project we will deploy in down the line.

scripts/purge.ps1
paulbalaji: good shout 👍

scripts/build.sh
paulbalaji: yes, essentially adding an extra layer in the build script sandwich for TC

jamiebrynes7: TC currently calls `build.sh` which has changed drastically and does different things.

I'd prefer to not need to modify TC builds while adding Buildkite functionality.

scripts/build.sh
```diff
@@ -24,12 +27,7 @@ LOG_FILE=${5:-} # Optional argument - will log to console otherwise
 
 ASSET_CACHE_ARG=${UNITY_ASSET_CACHE_IP:+-CacheServerIPAddress "${UNITY_ASSET_CACHE_IP}"}
 
-pushd "$(dirname "$0")/../"
-    source "scripts/pinned-tools.sh"
-    source "scripts/profiling.sh"
-
-    RUN_UNITY_PATH="$(pwd)/tools/RunUnity/RunUnity.csproj"
-popd
+RUN_UNITY_PATH="$(pwd)/tools/RunUnity/RunUnity.csproj"
```
paulbalaji: this needs to be updated

scripts/build.sh
paulbalaji: updated so TC --> build-test --> build-worker, buildkite --> pipeline --> build-worker

scripts/build-worker.sh
```diff
@@ -0,0 +1,28 @@
...
+  if [[ ${WORKER_TYPE} == "iOSClient" ]]; then
+      if ! isMacOS; then
+          echo "I can't build for iOS!"
+          exit 1
+      fi
+  fi
+
+  LOG_LOCATION="$(pwd)/../logs/${WORKER_TYPE}-${BUILD_TARGET}-${SCRIPTING_TYPE}.log"
+
+  scripts/build.sh "$(pwd)/../workers/unity" ${WORKER_TYPE} ${BUILD_TARGET} ${SCRIPTING_TYPE} "${LOG_LOCATION}"
```
jamiebrynes7, line 26: So there is an assumption about where the Unity project is. This is fine for now as they are all in the same place - but we may need to make this generic in the future.

Fine for this PR though

jamiebrynes7 (approved):
I think this is good

jessicafalk (approved):

PR #27, opened 2019-03-15
Log workaround

jamiebrynes7:
Adds a workaround so we can access logs without pinging EV. If there is an error, we cat the log file if the env variable is defined. 

Will test locally before merging


scripts/build-worker.sh
```diff
@@ -5,6 +5,16 @@ if [[ -n "${DEBUG-}" ]]; then
   set -x
 fi
 
+# Workaround until the artifact proxying service is set up properly.
```
jared-improbable, line 8: Link to ticket? :)

scripts/build-worker.sh
```diff
@@ -5,6 +5,16 @@ if [[ -n "${DEBUG-}" ]]; then
...
+# At least this gives us terminal output
+function printLog() {
+    if [[ ${LOG_LOCATION-} ]]; then
+        cat "${LOG_LOCATION}"
```
jared-improbable: This may be over-thinking, but would it be better to send to `stderr`?

jared-improbable (approved):

scripts/build-worker.sh
jamiebrynes7: Probably for correctness yes

scripts/build-worker.sh
jamiebrynes7: Will have a dig around and see if I can uncover it 😉

PR #33, opened 2019-05-02
Remove cd $projectRoot

oblm:
`cd $projectRoot` does not work as projectRoot is not defined.

You can see this command failing in CI [here](https://buildkite.com/improbable/gdk-for-unity-premerge/builds/1197#2f10817b-3180-4ada-8373-b38369da5f3a/204-210).

Thankfully, the purge script works anyway, it is seemingly directory agnostic. As such, I'm removing this command in order to keep things clean.


jamiebrynes7 (commented):
This is actually directory specific and the `projectRoot` should be passed in as an argument. I'm unsure why that's _not_ happening - https://github.com/spatialos/gdk-for-unity/blob/master/.buildkite/hooks/post-command as that looks fine to me.

The correct way to handle this either:
- Ask the caller of this script to CD to the correct directory
OR
- Validate that the argument is passed properly

oblm:
I'm closing this PR in favour of: [UTY-2012](https://improbableio.atlassian.net/browse/UTY-2012), a task that asks us to ensure **$projectRoot** is passed correctly.

@jamiebrynes7  is correct, **$projectRoot** is required to limit the scope of the **Kill-Dangling-Processes** function. If it is not called, we risk killin processes being used by builds running simultaneously on the same agent.

PR #38, opened 2019-07-02
Release tool changes

jamiebrynes7:
Through discussion, we landed on using a GitHub `gdk-robot` account for the release tool, then means that for the prep, we can use a fork-based workflow and for the actual release, it does everything short of publishing the release (requires small amount of manual work). 

- Removed references to attended mode. Designed to run only in CI. Can be run from your own account provided you create an access token and fork the repos.
- See comments in the `PrepCommand` file for a full description of the flow.
- Similarly, see comments in the `ReleaseCommand` for a full description.
- Changes to the underlying `GithubClient` and `GitClient` to remove unnecessary complexity and functionality after the changes to the workflow.


tools/ReleaseTool/GitClient.cs
```diff
@@ -16,78 +16,56 @@ internal class GitClient : IDisposable
...
         private const string GitCommand = "git";
         private const string PushArgumentsTemplate = "push {0} HEAD:refs/heads/{1}";
-        private const string FetchArguments = "fetch";
-        private const string PushForceFlag = " -f";
-        private const string SquashMergeArgumentsTemplate = "merge --squash {0} -m \"{1}\"";
-
+        private const string FetchArguments = "fetch {0}";
+        private const string CloneArgumentsTemplate = "clone {0} {1}";
+        private const string AddRemoteArgumentsTemplate = "remote add {0} {1}";
+
```
paulbalaji: nit: empty whitespace

paulbalaji: also Rider tells me you don't need `using CommandLine;` on line 4

tools/ReleaseTool/GitClient.cs
```diff
@@ -16,78 +16,56 @@ internal class GitClient : IDisposable
...
-        }
-
-        public void CheckoutRemoteBranch(string branch)
-        {
-            var branchRef = string.Format(RemoteBranchRefTemplate, branch, options.GitRemote);
+            var branchRef = string.Format(RemoteBranchRefTemplate, branch, remote ?? DefaultRemote);
             Logger.Info("Checking out branch... {0}", branchRef);
             Commands.Checkout(repo, branchRef);
         }
+
```
paulbalaji: nit: extra line

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -31,36 +30,33 @@ internal class PrepCommand
         private const string ChangeLogFilename = "CHANGELOG.md";
         private const string ChangeLogReleaseHeadingTemplate = "## `{0}` - {1:yyyy-MM-dd}";
         private const string ChangeLogUpdateGdkTemplate = "- Upgraded to GDK for Unity version `{0}`";
+        private const string ReleaseBranchNameTemplate = "feature/release-{0}";
 
         private const string PullRequestTemplate = "Release {0} - Pre-Validation";
-
+
```
paulbalaji: spare whitespace

tools/ReleaseTool/ReleaseCommand.cs
```diff
@@ -52,136 +41,107 @@ public ReleaseCommand(Options options)
...
+            }
 
-        private bool ShouldUpdateGdkVersion()
-        {
-            return !string.IsNullOrEmpty(options.GdkVersion);
+            return release;
         }
-
-        private Release CreateRelease(GitHubClient gitHubClient, Octokit.Repository gitHubRepo, GitClient gitClient)
+
```
paulbalaji: spare whitespace

tools/ReleaseTool/PrepCommand.cs
jamiebrynes7: I'm saving it for the winter

paulbalaji (approved):

tools/ReleaseTool/Common.cs
```diff
@@ -40,5 +44,21 @@ public static string AppendDirectorySeparator(string originalPath)
 
             return originalPath + directorySeparator;
         }
+
+        public class WorkingDirectoryScope : IDisposable
```
jared-improbable: Cute - split it out into its own file?

tools/ReleaseTool/Common.cs
```diff
@@ -40,5 +44,21 @@ public static string AppendDirectorySeparator(string originalPath)
...
+        {
+            private readonly string oldWorkingDirectory;
+            
+            public WorkingDirectoryScope(string newWorkingDirectory)
+            {
+                oldWorkingDirectory = Environment.CurrentDirectory;
+                Environment.CurrentDirectory = newWorkingDirectory;
```
jared-improbable: The setter for `CurrentDirectory` can throw a number of exceptions - how are you handling this?

tools/ReleaseTool/GitHubClient.cs
```diff
@@ -110,5 +93,23 @@ public ReleaseAsset AddAssetToRelease(Release release, string fileName, string c
...
+        {
+            if (!string.IsNullOrEmpty(options.GitHubToken))
+            {
+                octoClient.Credentials = new Credentials(options.GitHubToken);
+            }
+            else
+            {
+                if (!File.Exists(options.GitHubTokenFile))
+                {
+                    throw new ArgumentException("Failed to get GitHub Token as the file specified does not exist.");
```
jared-improbable: Print the file here

tools/ReleaseTool/Common.cs
jamiebrynes7: Hmmm I'm happy for the exceptions to propagate out here in the constructor since it means you've given this garbage data (and I'd otherwise do the exact same checks that the setter is doing). I.e. - pass the handling to the caller 

However, there's also the chance that it throws in the `Dispose()` (which would be odd as it meant the path from the original directory is either: null or empty, an invalid path, perms err). This could be thrown if the user removes the original directory while inside the scope, but not sure what the graceful thing to do is though.

We really shouldn't throw exceptions in disposes though..

tools/ReleaseTool/Common.cs
jamiebrynes7: Sure - I was actually a little surprised there wasn't something like this in the standard library, as its just `pushd` and `popd`.

tools/ReleaseTool/Common.cs
jamiebrynes7: Discussed offline - in this case, it may be best to let the exceptions bubble up since if these options fail its a full on failure case

PR #39, opened 2019-07-02
Update scripts to separate BUILD_TARGET from BUILD_ENVIRONMENT

paulbalaji:
Necessary for https://github.com/spatialos/gdk-for-unity/pull/1012

- add ability to pass in build target filters for a given build environment
- automatically build all targets for an environment, if environment variable is not specified
- combined `build-worker.sh` and `build.sh`
- renamed `prepare-unity-mobile.sh` to `prepare-unity-android.sh`
	- since this script's only job was to set up android dependencies
- removed some team city specific things
	- no more `profiling.sh`
	- no more `prepare-unity.sh`
	- removed some `if buildkite` checks


scripts/build.sh
```diff
@@ -2,7 +2,7 @@
 
 # This is a generic build script for the GDK for Unity.
 # The expected usage is as follows:
-#   bash build.sh <unity_project_dir> <worker_type> <build_target> <scripting_backend> <log_file (optional)>
+#   bash build.sh <unity_project_dir> <worker_type> <build_environment> <build_target_filter> <scripting_backend> <log_file (optional)>
```
paulbalaji: need to figure out how optional arguments _actually_ work in bash

scripts/build-worker.sh
```diff
@@ -1,6 +1,16 @@
 #!/usr/bin/env bash
 set -e -u -o pipefail
 
+# This is a generic build script for the GDK for Unity.
+#
+# Expected environment variables:
+#   WORKER_TYPE
+#   BUILD_ENVIRONMENT
+#   SCRIPTING_TYPE
```
paulbalaji:
```suggestion
#   SCRIPTING_BACKEND
```

scripts/build-worker.sh
```diff
@@ -10,29 +20,82 @@ fi
...
     if [[ ${LOG_LOCATION-} ]]; then
-        cat "${LOG_LOCATION}" 1>&2 
+        cat "${LOG_LOCATION}" 1>&2
     fi
 }
 
 trap printLog ERR
 
 source "$(dirname "$0")/pinned-tools.sh"
+source "$(dirname "$0")/profiling.sh"
```
jamiebrynes7: What do we need from this?

jamiebrynes7: Actually the `markStartOfBlock`, etc things are TC only. We can kill all that too 😄

scripts/build-worker.sh
```diff
@@ -10,29 +20,82 @@ fi
...
 
-echo "Building for: ${WORKER_TYPE} ${BUILD_TARGET} ${SCRIPTING_TYPE}"
+echo "Building for: ${WORKER_TYPE} ${BUILD_ENVIRONMENT} ${SCRIPTING_BACKEND}"
 
 pushd "$(dirname "$0")/../"
-  if [ ${WORKER_TYPE} == "AndroidClient" ] || [ ${WORKER_TYPE} == "MobileClient" ]; then
-      scripts/prepare-unity-mobile.sh "$(pwd)/../logs/PrepareUnityMobile.log"
-  fi
+    if [ "${WORKER_TYPE}" == "AndroidClient" ] || [ "${WORKER_TYPE}" == "MobileClient" ]; then
```
jamiebrynes7: Hmm is the first part of the check required anymore? No old PRs will be running through this I don't think?

scripts/build-worker.sh
```diff
@@ -10,29 +20,84 @@ fi
...
+        if [[ ${BUILD_TARGET_FILTER-} != "ios" ]]; then
```
jamiebrynes7, line 35: What if I do `ios,android` 😛

scripts/build-worker.sh
```diff
@@ -10,29 +20,84 @@ fi
...
+            scripts/prepare-unity-android.sh "$(pwd)/../logs/PrepareUnityMobile.log"
+        fi
+    fi
+
+    if [[ ${WORKER_TYPE} == "iOSClient" ]]; then
+        if ! isMacOS; then
+            echo "I can't build for iOS!"
```
jamiebrynes7: `Oh noes`

scripts/build-worker.sh
```diff
@@ -10,29 +20,84 @@ fi
...
+    #       ASSET_CACHE_ARG="-CacheServerIPAddress ${UNITY_ASSET_CACHE_IP}"
+    #   else
+    #       ASSET_CACHE_ARG=""
+    #   fi
+    ASSET_CACHE_ARG=${UNITY_ASSET_CACHE_IP:+-CacheServerIPAddress "${UNITY_ASSET_CACHE_IP}"}
+
+    RUN_UNITY_PATH="$(pwd)/tools/RunUnity/RunUnity.csproj"
+
+    pushd "$(pwd)/../workers/unity"
+        rm -rf ./build/worker/
```
jamiebrynes7: Hmmm I wonder why we did this & is it required in BK land

scripts/build-worker.sh
```diff
@@ -10,29 +20,84 @@ fi
...
+
+        if [[ -n ${BUILD_TARGET_FILTER-} ]]; then
+            BLOCK_MESSAGE="Building ${WORKER_TYPE} for ${BUILD_ENVIRONMENT} on ${BUILD_TARGET_FILTER} using ${SCRIPTING_BACKEND}"
+            LOG_FILE="$(pwd)/../logs/${WORKER_TYPE}-${BUILD_ENVIRONMENT}-${BUILD_TARGET_FILTER}-${SCRIPTING_BACKEND}.log"
+
+            markStartOfBlock "${BLOCK_MESSAGE}"
+
+            dotnet run -p "${RUN_UNITY_PATH}" -- \
```
jamiebrynes7: You can pull this out of the if-else. 

Define all the variables (for the buildTargetFilter you can use the same trick as the asset cache argument above

Then if/when we change this we only need to change it in once place

scripts/build-worker.sh
paulbalaji: that's fine because we'd still need to setup dependencies for the android build

I believe using `*"ios"*` does the maybe-troll substring matching?

scripts/build-worker.sh
paulbalaji: https://github.com/spatialos/gdk-for-unity-shared-ci/pull/13/commits/e099c93243af35974200acc44f387fdfde556cec

think I added it as a workaround for some rogue team city bug

probably not needed in BK

jamiebrynes7 (approved):

PR #47, opened 2019-07-18
Last Fix I think

jamiebrynes7:


zeroZshadow (approved):

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -143,6 +145,14 @@ public int Run()
             return 0;
         }
 
+        private void UpdateManifestJson(GitClient gitClient)
+        {
+            var manifestFile = Path.Combine(Directory.GetCurrentDirectory(), "workers", "unity", "Packages",
+                ManifestJsonFilename);
+            
+            UpdatePackageJson(manifestFile, gitClient);
```
zeroZshadow, line 154: Can this code actually be reused?

jamiebrynes7: It can now

tools/ReleaseTool/PrepCommand.cs
```diff
@@ -165,23 +175,27 @@ private void UpdatePackageJson(string packageFile, GitClient gitClient)
...
-
+                
                 if (jsonObject.ContainsKey(PackageJsonDependenciesString))
                 {
                     var dependencies = (JObject) jsonObject[PackageJsonDependenciesString];
 
                     foreach (var property in dependencies.Properties())
                     {
-                        if (property.Name.StartsWith(PackageJsonDependenciesPrefix))
+                        // If its an Improbable package and its not a file: reference.
```
paulbalaji:
```suggestion
                        // If it's an Improbable package and it's not a "file:" reference.
```

paulbalaji (approved):

PR #56, opened 2019-08-23
Unity 2019.2.0f1 upgrade

zeroZshadow:
This update removes the Android preparation step, which is no longer required since 2019.2
It also fixes the script that scrapes the required unity version from the project.


scripts/build-worker.sh
```diff
@@ -18,12 +18,6 @@ fi
 source "$(dirname "$0")/pinned-tools.sh"
 
 pushd "$(dirname "$0")/../"
-    if [[ "${WORKER_TYPE}" == "MobileClient" ]]; then
```
jamiebrynes7, line 21: 🎉

jamiebrynes7 (approved):
Looks good. Lets get the blank and FPS upgrade PRs in place before we merge this :)

tools/UnityPaths/UnityPaths.cs
```diff
@@ -65,7 +83,7 @@ public static string TryGetUnityPath()
             }
 
             throw new Exception(
-                $"Could not find Unity in\n  {UnityHomeEnvVar}={UnityHomeValue}\n  {improbableUnityPath}\n  {hubPath}\n  {DefaultUnityPath}");
+                $"Could not find Unity in\n  [{improbableUnityPath}]\n  [{hubPath}]\n  [{DefaultUnityPath}]");
```
jessicafalk:
```suggestion
                $"Could not find Unity in\n[{improbableUnityPath}]\n[{hubPath}]\n[{DefaultUnityPath}]");
```

nit: spaces are not really needed?

jessicafalk (approved):

tools/UnityPaths/UnityPaths.cs
zeroZshadow: It just indents the paths in the log files, rather unneeded.

zeroZshadow: Will do

PR #59, opened 2019-09-26
Enabling XCode build

jessicafalk:
Adding a step to build the XCode project, if we build for iOS


scripts/build-worker.sh
```diff
@@ -32,13 +32,17 @@ pushd "$(dirname "$0")/../"
...
-        BUILD_TARGET_FILTER_ARG="+buildTargetFilter ${BUILD_TARGET_FILTER}"
+        BUILD_TARGET_FILTER_ARG="+buildTargetFilter ${BUILD_TARGET_FILTER} "
     else
         BLOCK_MESSAGE="Building ${WORKER_TYPE} for ${BUILD_ENVIRONMENT} using ${SCRIPTING_BACKEND}"
         LOG_FILE="$(pwd)/../logs/${WORKER_TYPE}-${BUILD_ENVIRONMENT}-${SCRIPTING_BACKEND}.log"
         BUILD_TARGET_FILTER_ARG=""
     fi
 
+    if [[ -n ${TARGET_IOS_SDK-} ]]; then
+        BUILD_TARGET_FILTER_ARG+="+targetiOSSdk ${TARGET_IOS_SDK}"
```
jamiebrynes7: Shouldn't this be its own argument? Rather than tagging onto `BUILD_TARGET_FILTER_ARG` ?

zeroZshadow: Should this use a different variable? Instead of reusing the filter

scripts/build-worker.sh
```diff
@@ -55,5 +59,17 @@ pushd "$(dirname "$0")/../"
             +buildEnvironment "${BUILD_ENVIRONMENT}" \
             +scriptingBackend "${SCRIPTING_BACKEND}" \
             "${BUILD_TARGET_FILTER_ARG}"
+
+        if [[ "${BUILD_TARGET_FILTER}" == "ios" ]]; then
```
jamiebrynes7: What if `BUILD_TARGET_FILTER` is `ios android` 😉

scripts/build-worker.sh
```diff
@@ -55,5 +59,17 @@ pushd "$(dirname "$0")/../"
...
+            echo "--- Building XCode Project :hammer_and_wrench:"
+
+            dotnet run -p "${RUN_UNITY_PATH}" -- \
+                -projectPath "." \
+                -batchmode \
+                -quit \
+                -logfile "${LOG_FILE}" \
```
jamiebrynes7: Should be in its own log file

scripts/build-worker.sh
```diff
@@ -55,5 +59,17 @@ pushd "$(dirname "$0")/../"
...
+                -executeMethod "Improbable.Gdk.Mobile.iOSLaunchUtils.Build" \
```
jamiebrynes7: Why is the `Build` method in `iOSLaunchUtils` 🤔

jamiebrynes7 (changes requested):

scripts/build-worker.sh
zeroZshadow: Does this work if `BUILD_TARGET_FILTER` is not set?

scripts/build-worker.sh
jessicafalk: cause you first need to build the XCode project, before you can  launch it. 
Happy to change the name if you have a better suggestion

zeroZshadow (approved):

scripts/build-worker.sh
jamiebrynes7: `Improbable.Gdk.Mobile.XcodeUtils.Build` and we can throw all the random stuff we need to hack in for XCode in there

scripts/build-worker.sh
```diff
@@ -54,6 +60,19 @@ pushd "$(dirname "$0")/../"
...
             "${BUILD_TARGET_FILTER_ARG}"
+
+        if [[ "${BUILD_TARGET_FILTER-}" =~ "ios" ]]; then
+            echo "--- Building XCode Project :hammer_and_wrench:"
+
+            dotnet run -p "${RUN_UNITY_PATH}" -- \
+                -projectPath "." \
+                -batchmode \
+                -quit \
+                -logfile "$(pwd)/../logs/${WORKER_TYPE}-xcode-build.log" \
```
jamiebrynes7:
```suggestion
                -logfile "$(pwd)/../logs/${WORKER_TYPE}-${BUILD_ENVIRONMENT}-xcode-build.log" \
```

Prevent contention between local and cloud :)

scripts/build-worker.sh
```diff
@@ -54,6 +60,19 @@ pushd "$(dirname "$0")/../"
             +buildWorkerTypes "${WORKER_TYPE}" \
             +buildEnvironment "${BUILD_ENVIRONMENT}" \
             +scriptingBackend "${SCRIPTING_BACKEND}" \
+            "${TARGET_IOS_SDK_ARG}" \
             "${BUILD_TARGET_FILTER_ARG}"
+
+        if [[ "${BUILD_TARGET_FILTER-}" =~ "ios" ]]; then
+            echo "--- Building XCode Project :hammer_and_wrench:"
```
jamiebrynes7:
```suggestion
            echo "--- Building XCode Project :xcode:"
```

Can't believe you wouldn't check for a custom emoji 😞

jamiebrynes7 (approved):

scripts/build-worker.sh
jessicafalk: :O How could I not have done that

scripts/build-worker.sh
```diff
@@ -64,7 +64,7 @@ pushd "$(dirname "$0")/../"
             "${BUILD_TARGET_FILTER_ARG}"
 
         if [[ "${BUILD_TARGET_FILTER-}" =~ "ios" ]]; then
-            echo "--- Building XCode Project :hammer_and_wrench:"
+            echo "--- Building XCode Project :xcode:"
```
paulbalaji, line 67: 💯 was just about to suggest this

paulbalaji (approved):

PR #72, opened 2020-05-28
Accelerator

paulbalaji:
Add function to return arguments for Unity Accelerator, if `ACCELERATOR_ENDPOINT` environment variable is provided.


scripts/pinned-tools.sh
```diff
@@ -83,3 +83,7 @@ function isDocsBranch() {
   fi
   return 1
 }
+
+function getAcceleratorArgs() {
+  echo ${ACCELERATOR_ENDPOINT:+-adb2 -enableCacheServer -cacheServerEnableDownload true -cacheServerEnableUpload true -cacheServerEndpoint "${ACCELERATOR_ENDPOINT}"}
```
paulbalaji, line 88: I spent a day and a half figuring out why nothing was working until I stumbled upon a forum post where someone says to add `-adb2` if trying to access Accelerator in batch mode

:tableflip:

zeroZshadow: Could it be some of our projects have not been converted to asset db 2?

paulbalaji: I observed this on all projects

jamiebrynes7: We did do the asset db upgrade & then a full asset refresh during the 2019.3 upgrade though.

paulbalaji: Fresh clones also show asset db 2 is used in the preferences

jamiebrynes7 (approved):

zeroZshadow (approved):

PR #74, opened 2020-06-05
New Buildkite instrumentation format

paulbalaji:
As per the guide in https://improbableio.atlassian.net/browse/UTY-2600


paulbalaji:
To be replaced by V2

PR #76, opened 2020-06-25
Add secrets to publishing Docker image for China NPM registry

jamiebrynes7:
There will be a follow up PR in `gdk-for-unity` to add the 'publish' step.


ci/publish-packages.sh
```diff
@@ -9,7 +9,7 @@ set -e -u -o pipefail
 function getSecrets() {
     if [[ ${BUILDKITE:-} ]]; then
 
-        export SPATIAL_OAUTH_DIR=$(mktemp -d)
+        export SPATIAL_OAUTH_DIR=$(mktemp -d "${TMPDIR:-/tmp}/XXXXXXXXX")
```
jamiebrynes7, line 12: These changes were for iterating on a MacOS machine, this still works on the Linux agents, but it just makes it easier in the future :)

ci/publish-packages.sh
```diff
@@ -35,6 +35,22 @@ function getSecrets() {
...
+        # TODO: Read secret and setup .npmrc file to mount into the docker container
+        imp-ci secrets read \
+            --environment="production" \
+            --secret-type="generic-token" \
+            --buildkite-org="improbable" \
+            --secret-name="unity/npm-spatialos-china-dogbot" \
+            --field="token" \
+            --write-to="${NPMRC_FILE}"
+
+        # We don't want to write the API key to the console, so ewe use sed to append the contents to the front of the file to
```
paulbalaji:
```suggestion
        # We don't want to write the API key to the console, so we use sed to append the contents to the front of the file to
```

zeroZshadow (approved):

ci/publish-packages.sh
jamiebrynes7: <details>
<summary>But ewe is using sed!</summary>

image

</details>

paulbalaji (approved):

PR #80, opened 2020-07-29
UTY-2646: update dependencies.pinned

BryanJY-Wong:
Uploading macos installer to gcs bucket

### DIFF
--- a/ci/docker/release-tool.Dockerfile
+++ b/ci/docker/release-tool.Dockerfile
@@ -1,4 +1,4 @@
-FROM microsoft/dotnet:2.2-sdk as build
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1 as build
 
 # Copy everything and build
 WORKDIR /app
@@ -6,7 +6,7 @@ COPY ./tools ./
 RUN dotnet publish -c Release -o out
 
 # Build runtime image
-FROM mcr.microsoft.com/dotnet/core/runtime:2.2
+FROM mcr.microsoft.com/dotnet/core/runtime:3.1
 WORKDIR /app
 COPY --from=build /app/*/out ./
 
--- a/ci/docker/release-tool.Dockerfile
+++ b/ci/docker/release-tool.Dockerfile
@@ -8,7 +8,7 @@ RUN dotnet publish -c Release -o out
 # Build runtime image
 FROM mcr.microsoft.com/dotnet/core/runtime:3.1
 WORKDIR /app
-COPY --from=build /app/*/out ./
+COPY --from=build /app/out ./
 
 # Setup GIT
 RUN apt-get update && \
--- a/ci/docker/release-tool.Dockerfile
+++ b/ci/docker/release-tool.Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/dotnet/core/sdk:3.1 as build
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1-alpine as build
 
 # Copy everything and build
 WORKDIR /app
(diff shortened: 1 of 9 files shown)


ci/docker/release-tool.Dockerfile
```diff
@@ -1,14 +1,14 @@
-FROM microsoft/dotnet:2.2-sdk as build
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1-bionic as build
```
zeroZshadow: Why is this specifically using the bionic image?

BryanJY-Wong: Because the regular one doesn't work 😢; the Git library fails to find one of the binaries and somehow it only works on this image.

zeroZshadow: What specifically broke? And how did you find this image as the solution?

BryanJY-Wong: https://github.com/libgit2/libgit2sharp/issues/1703
I tried using alpine, but it didn't work.

zeroZshadow: This bionic image would only be required in the step that runs in then? This step likely wont need it.

ci/docker/release-tool.Dockerfile
```diff
@@ -1,14 +1,14 @@
...
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1 as build
 
 # Copy everything and build
 WORKDIR /app
 COPY ./tools ./
 RUN dotnet publish -c Release -o out
 
 # Build runtime image
-FROM mcr.microsoft.com/dotnet/core/runtime:2.2
+FROM mcr.microsoft.com/dotnet/core/runtime:3.1-bionic
```
jamiebrynes7, line 10:
```suggestion
# Requires the bionic image due to runtime requirements of the libgit2sharp library
FROM mcr.microsoft.com/dotnet/core/runtime:3.1-bionic
```

tools/ReleaseTool.Tests/ReleaseTool.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFramework>netcoreapp2.1</TargetFramework>
+        <TargetFramework>netcoreapp3.1</TargetFramework>
```
jamiebrynes7, line 4: @zeroZshadow do we want to pin the lang version in all of these too?

zeroZshadow: Would be nice. I already saw that some were set to 7.3 and some 7.2

paulbalaji: +1 to pinning lang versions

BryanJY-Wong: which version?

paulbalaji: Could stay on 7.3? If we need C#8 features we can update as/when needed?

jamiebrynes7: Equally 8 is fully backwards compatible with 7.3, so its not like choosing 8 has any downsides for us

jamiebrynes7 (approved):

tools/ReleaseTool.Tests/ReleaseTool.Tests.csproj
paulbalaji: Yeah that's true

paulbalaji (approved):

